PR #41, opened 2019-10-31
Feature/multi thresholding

auphelia:
Added multi-thresholding stand alone function and made it executable by adding an execute custom node function and modifying the onnx_exec.py to decide on the domain value of each node if the node should be executed by onnxruntime or with the new execution function. Also added unit tests for multi-thresholding, execute custom node and one unit tests that creates a graph with a custom and a standard node and runs onnx_exec.py.


maltanar:
Thank you!

PR #44, opened 2020-02-13
[fix]  batchnorm transform always removes producer and consumer nodes

### LINKED ISSUE #43: Streamline input name mismatch in transformed matmul nodes
The Streamline transform generates strange names in inputs of the transformed MatMul nodes that are actually missing in the graph causing a disconnected network. 
For instance the following code generate a disconnected graph:
```python
    lfc = LFC(weight_bit_width=1, act_bit_width=1, in_bit_width=1)
    checkpoint = torch.load(trained_lfc_checkpoint, map_location="cpu")
    lfc.load_state_dict(checkpoint["state_dict"])
    bo.export_finn_onnx(lfc, (1, 1, 28, 28), export_onnx_path)
    model = ModelWrapper(export_onnx_path)
    model = model.transform(InferShapes())
    model = model.transform(FoldConstants())
    model = model.transform(GiveUniqueNodeNames())
    model = model.transform(GiveReadableTensorNames())
    model = model.transform(Streamline())
```
the corresponding net shown in netron is:
image

The problem seems to happen in BatchNormToAffine()

AndreaRigoni:
I think the intent of removing producer and consumer was limited to the squeeze and unsqueeze nodes but in reality the function was removing them in any case.

this should fix #43

### DIFF
--- a/src/finn/transformation/batchnorm_to_affine.py
+++ b/src/finn/transformation/batchnorm_to_affine.py
@@ -67,8 +67,10 @@ class BatchNormToAffine(Transformation):
                 # remove old nodes
                 graph.node.remove(n)
                 if consumer is not None:
-                    graph.node.remove(consumer)
+                    if consumer.op_type == "Squeeze":
+                        graph.node.remove(consumer)
                 if producer is not None:
-                    graph.node.remove(producer)
+                    if producer.op_type == "Unsqueeze":
+                        graph.node.remove(producer)
         model = model.transform(InferShapes())
         return (model, graph_modified)


maltanar:
Thanks Andrea! The fix itself looks fine, would you mind adding an additional test case (BatchNorm surrounded by non-(un)Squeeze nodes) to tests/transformation/test_batchnorm_to_affine.py that checks for the correct behavior?

AndreaRigoni:
Hi Yaman, Actually the fix is implicitly tested in test_convert_to_hls_layers.py because the TFC model has the batchnorm layer but not the squeeze. (as far as I understood)
Have you a test CI somewhere?

maltanar:
Screenshot from 2020-02-17 15-05-30

TFC also has unsqueeze-squeeze nodes around BatchNorm, so the behavior this fix is targeting would not be tested as part of `test_convert_to_hls_layers`. For me that particular test already succeeds without this fix, so something else may be going on here.

We do have a test suite that you can launch via `python setup.py test` from the finn root dir. If you'd like to execute a single test (since the whole test suite takes quite long) you can do `python setup.py test --addopts "-k test_name"`.  @auphelia is currently looking into adding Jenkins.

AndreaRigoni:
It seems that (probably because of my onnx or torch versions) I have the squeeze un-squeeze around the batchnorm removed at a certain time..
image

maltanar:
I found a testcase that we already have (CNV-w1a1) that reproduces the problem, and your fix also fixes that one. Thanks for your contribution!

PR #63, opened 2020-03-26
Update getting_started.rst

RSPwFPGAs:
On Ubuntu 18.04 with Docker newly installed, the docker group should be created and your current user should be added to the docker group, so your current user can access the Unix socket that the Docker daemon creates .


maltanar:
Thanks @RSPwFPGAs -- could you make the pull request to the `dev` branch and not the master branch?

PR #66, opened 2020-04-03
Docker post install without sudo

RSPwFPGAs:


RSPwFPGAs:
Hi maltanar, I will let you add the modified content of 411e897 in a later commit. I just cannot figure out how to add it cleanly.

PR #67, opened 2020-04-07
Deal gracefully with missing PyVerilator

volcacius:
PyVerilator is not supported on Windows. I think, while it makes sense to keep it as a required dependency, it would be good to be able to deal with its absence in an ordered manner. 
This is required to support Brevitas-FINN tests on Windows.


maltanar:
Thanks! I should point out that we won't be able to support a full FINN flow on Windows as we are dependent on several other things that for the moment are Linux-only, but the part that's required to run the test_brevitas* tests has much fewer dependencies and should be OK.

PR #71, opened 2020-04-17
Feature/check topological order

### LINKED ISSUE #61: Do topological sort checking prior to execution
Although the ONNX standard requires `model.graph.node` to be topologically sorted, there is nothing in the data structures themselves to restrict this. FINN assumes that graphs *are* topologically sorted in e.g. `execute_onnx`. Execution of a graph without topological sorting will produce incorrect results without any error message. Instead, we should have an analysis pass that checks for topological sorting, which gets called prior to execution (and anywhere else that depends on topological sorting for correct functioning).

auphelia:
Add analysis pass to check if `model.graph.node` is topologically sorted and unit test for this analysis pass. 
Closes #61


maltanar:
Thanks @auphelia -- it's a good start but the analysis pass should not be limited to linear graphs. https://stackoverflow.com/a/54995143/1520469 (under "vertex approach") is what I would go for in terms of the implementation. I'd suggest the following steps:

1. The current `ModelWrapper.find_consumer` function only returns one consumer at most. Add a new function `ModelWrapper.find_consumers` that returns a list of consumer nodes for a given tensor name. Then add a `ModelWrapper.find_successors` function that returns a list of all successors (all consumer nodes for all outputs) given a node.

2. In the analysis pass, iterate over the sequence of `model.graph.node`s, keeping track of the current index in the node sequence. Call `find_successors` for each node. Find the indices for each node in the returned successor list. All successors must have index > current node index, else return false.

3. Add at least four test graphs (can be small, manually constructed ones), both linear and nonlinear, both with expected return value True and False.

maltanar:
Thanks -- I've changed the name of the analysis pass to `nodes_topologically_sorted` to give a bit more info to the user, but otherwise looks good.

PR #78, opened 2020-04-22
Support for node-local transformations

HenniOVP:
Hi,
here is my current implementation for the `NodeLocalTransformation` and the updated `HLSSynth_IPGen`, which now subclasses `NodeLocalTransformation`.

By default all operations still run in serial, because I have noticed that there is a case where the synthesis will overshoot the 16 GB memory mark when working on the LFC network. And this already happens for only two threads. See this plot:
memory_spike_slide
On some systems, as my desktop system this is not a problem, but my laptop would then forcefully kill the vivado process, which caused errors later down the line.

So for now the default is set such that one needs to consciously enable parallelization.

### DIFF
--- a/src/finn/transformation/__init__.py
+++ b/src/finn/transformation/__init__.py
@@ -48,6 +48,7 @@ Guide to writing FINN transformations
 """
 
 from abc import ABC, abstractmethod
+import multiprocessing as mp
 
 
 class Transformation(ABC):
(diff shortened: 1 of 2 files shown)


maltanar:
@HenniOVP I'm just having a quick look and trying it out myself now, looks good! I agree that using NUM_DEFAULT_WORKERS=1 is a good idea as a safe default.

I'm going to introduce an environment variable NUM_DEFAULT_WORKERS, add an utility function to read its value (with a default of 1) and slightly change the behavior of the constructor parameter so that None uses the env.var. value, and 0 sets it to the number of CPU cores. This should give the users a quick way to set this up as part of their environment without having to modify the code -- and if they do want per-instance parallelization for transformations this is still possible.

PR #79, opened 2020-04-22
Feature/decoupled mode large matrix sizes

auphelia:
Enable StreamingFCLayer decoupled mode for larger matrix sizes.
Write weight stream in multiple .dat files and set mem stream depth higher. Additional change in finn-rtllib (ramb18_wf_dualport.v) to enable simulation with PyVerilator. Add rtlsim test for StreamingFCLayer_Batch MW=MH=128 in decoupled mode to verify functionality.


maltanar:
I wanted to test decoupled mode on even larger matrices so I changed `mem_mode="decoupled"` in `test_end2end_cnv_w1a1` -- the good news is that IP generation is substantially faster compared to `mem_mode="const"` but during the `test_end2end_cnv_w1a1_verify_dataflow_part` step the test fails as follows:

```
tests/end2end/test_end2end_cnv_w1a1.py::test_end2end_cnv_w1a1_verify_dataflow_part Fatal Python error: Aborted

Current thread 0x00007fa17799e700 (most recent call first):
  File "/workspace/pyverilator/pyverilator/pyverilator.py", line 693 in eval
  File "/workspace/pyverilator/pyverilator/pyverilator.py", line 668 in _post_write_hook
  File "/workspace/pyverilator/pyverilator/pyverilator.py", line 650 in _write_32
  File "/workspace/pyverilator/pyverilator/pyverilator.py", line 644 in _write
  File "/workspace/pyverilator/pyverilator/pyverilator.py", line 676 in _sim_init
  File "/workspace/pyverilator/pyverilator/pyverilator.py", line 514 in __init__
  File "/workspace/pyverilator/pyverilator/pyverilator.py", line 486 in build
  File "/workspace/finn/src/finn/custom_op/fpgadataflow/streamingfclayer_batch.py", line 676 in execute_node
```
I think you mentioned having this error before while working on the `$readmemh` fix for large matrix sizes. Could you have a look @auphelia ?

maltanar:
Just adding this comment here to summarize what the problems were and what needed to be fixed, as there were a few and some of them were tricky.

0. The MEM_DEPTH parameter for `memstream.v` was set incorrectly, should be equal to WMEM rounded up to nearest 1024 (8e22e2c)
1. The STREAM_DEPTH parameter for `memstream.v` was set incorrectly, should be equal to WMEM and not rounded up to nearest 1024 (6ea013d)
2. The "programmable full" FIFO was not properly implemented for the FIFO between the weight streamer and the MVAU  (33559c1 and 73e51c7)
3. Vivado was not happy with the `$sformatf` call in the updated `ram18_wf_dualport.v` so now using `ifdef`s to keep both Vivado and Verilator happy (41b9ffa)

With all this in place, `test_end2end_tfc_w1a1` succeeds all the way to hardware using `mem_mode=decoupled` and the own test `test_fpgadataflow_fclayer_large_depth_decoupled_mode` is also still passing. This is merged into dev and the default mode for `test_end2end_tfc_w1a1` is now `mem_mode=decoupled`.

PR #85, opened 2020-04-29
Feature/streaming fifo

auphelia:
Add new fpgadataflow node StreamingFIFO and transformation to insert it automatically with the correct depths in the model.


tests/end2end/test_end2end_tfc_w1a1.py
```diff
@@ -199,7 +204,7 @@ def test_end2end_tfc_w1a1_verify_dataflow_part():
     ret_rtlsim_whole = execute_onnx(model, inp_dict, True)
     res_rtlsim_whole = ret_rtlsim_whole[out_name]
     assert np.isclose(res_npysim, res_rtlsim_nodebynode).all()
-    assert np.isclose(res_npysim, res_rtlsim_whole).all()
+    assert np.isclose(res_rtlsim_nodebynode, res_rtlsim_whole).all()
```
maltanar, line 207: any particular reason why the comparison here changed?

PR #87, opened 2020-04-30
Feature/graph order util fct

auphelia:
Add modelwrapper functions to get a list of consumers of a given tensor (find_consumers), a list of successors (find_successors) and predecessors (find_predecessors) of a given node. Add a modelwrapper function to return the current node index of a given node. Integrate unit tests into test_modelwrapper.py.

### DIFF
--- a/src/finn/core/modelwrapper.py
+++ b/src/finn/core/modelwrapper.py
@@ -301,6 +301,27 @@ class ModelWrapper:
         except ValueError:
             return None
 
+    def find_successors(self, node):
+        successors = []
+        try:
+            for outp_tensor in node.output:
+                tensor_consumer_list = self.find_consumers(outp_tensor)
+                for consumer in tensor_consumer_list:
+                    successors.append(consumer)
+            return successors
+        except ValueError:
+            return None
+
+    def find_predecessors(self, node):
+        predecessors = []
+        try:
+            for inp_tensor in node.input:
+                producer = self.find_producer(inp_tensor)
+                predecessors.append(producer)
+            return predecessors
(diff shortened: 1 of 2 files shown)


src/finn/core/modelwrapper.py
```diff
@@ -288,6 +288,46 @@ def find_consumer(self, tensor_name):
...
+        for n in self._model_proto.graph.node:
+            for inp_tensor in n.input:
+                if inp_tensor == tensor_name:
+                    consumers.append(n)
+        if consumers != []:
+            return consumers
+        else:
+            return None
+
+    def find_successors(self, node):
```
maltanar, line 304: either rename to find_direct_successors (my recommendation) or implement actual find_successors -- in graph theory, a successor is *any* node reachable "downstream" from this node, so if you have a graph x -> y -> z -> t the successors of x are y, z, y. "direct successor" would be a better description of the current behavior.

src/finn/core/modelwrapper.py
```diff
@@ -288,6 +288,46 @@ def find_consumer(self, tensor_name):
...
+            tensor_consumer_list = self.find_consumers(outp_tensor)
+            if tensor_consumer_list is not None:
+                for consumer in tensor_consumer_list:
+                    successors.append(consumer)
+        if successors != []:
+            return successors
+        else:
+            return None
+
+    def find_predecessors(self, node):
```
maltanar, line 318: same as above, rename to find_direct_predecessors

maltanar:
Looking good, thanks @auphelia !

Issue #77, opened 2020-04-20
Add HLSCustomOp for standalone ReLU activation layers


maltanar:
Add a `HLSCustomOp` type for the finn-hlslib function `ReLU_Batch`.

PR #90, opened 2020-05-05
Implement HLS Custom Op for standalone thresholding

quetric:
This implements a hls custom op that maps directly to multithresholds. It is useful when multithresholds cannot be fused into MVAUs. 

Generally, this op can implement any activation function including ReLU so this PR affects #77


quetric:
Closing this, will merge multiple features into a single PR

maltanar:
I would actually prefer several smaller PRs instead of one large one.

Issue #89, opened 2020-05-05
rtlsim fails when using Vivado 2019.2


quetric:
How to reproduce:

```
sh run-docker.sh
python setup.py test --addopts "-k test_fpgadataflow_fclayer_rtlsim --pdb"
```

Expected behaviour: test passes

Actual behaviour: test fails with an error in the pyverilator compile stage:
```
%Error: /tmp/finn_dev_lpetrica/code_gen_ipgen_StreamingFCLayer_Batch_0_vew52yqa/project_StreamingFCLayer_Batch_0/sol1/impl/verilog/StreamingFCLayer_Batch_0_StreamingFCLayer_Batch_0.v:118: Cannot find file containing module: regslice_both
```
Note that in contrast to 2019.1 and earlier, 2019.2 puts regslice modules in a separate verilog file.

quetric:
Debugged this further. Problem is caused by the naming of the verilog file containing regslice modules, which starting in 2019.2, is called `regslice_core.v`. Verilator assumes modules are found in (System)Verilog files of the same name, which is not the case here. 

The solution is to pass both the top Verilog file and `regslice_core.v` to verilator along with a `--top-module` argument indicating the name of the top module. However, PyVerilator does not support either passing of `--top-module` or providing multiple verilog files to Verilator.

quetric:
Submitted issue to PyVerilator: https://github.com/csail-csg/pyverilator/issues/10

maltanar:
Thanks for finding and fixing this @quetric. Changes have been merged into our fork of PyVerilator and I've updated `dev` to use the latest version in 5ab626ede5137df954a07c7c9989393170c6cb9e -- hopefully this fixes the issues you have observed with Vivado 2019.2

Issue #75, opened 2020-04-20
Transformation(s) to insert top-k nodes at the end of network


maltanar:
FINN should have a `Transformation` that inserts an [ONNX `TopK` node](https://github.com/onnx/onnx/blob/master/docs/Operators.md#topk) at the end of a classification network. This will pave the way for doing the top-K selection operation in hardware and reducing the output bandwidth requirements. Once #74 is resolved, the ONNX TopK node can be converted into the HLSCustomOp equivalent as part of a new `ConvertToHLS` transformation, and be synthesized down to hardware.

maltanar:
@quetric expressed interest in working on this, but I couldn't assign him to this issue for some reason. Adding this comment here so that other interested parties can track the progress.

quetric:
Fixed in #84

maltanar:
#84 adds the transformation to insert the TopK node into the graph, we'll still need another `ConvertToHLSLayers`-type transformation that converts the `TopK` to `LabelSelect_Batch` once #74 is resolved.

Issue #74, opened 2020-04-20
Add HLSCustomOp for LabelSelect_Batch


maltanar:
finn-hlslib has a [LabelSelect_Batch](https://github.com/Xilinx/finn-hlslib/blob/e90374ffc4cd440cbeb67451e5f26d2123b56461/maxpool.h#L322) class that can extract the indices of top-k entries from an input vector, which is useful for classification networks. The corresponding `HLSCustomOp` should be supported in FINN.

PR #97, opened 2020-05-12
Multiple issues fixed

quetric:
Fixes #90 #74 #75 #89 and adds HLS Custom OP for elementwise addition plus large speed-up of Multithreshold. It was quite a large merge with the latest updates to dev, so some careful review is required for files which aren't new.


maltanar:
This needs to be broken down into separate PRs, it's too many changes all across the place to merge all at once.

PR #103, opened 2020-05-14
Feature/vectorized multithreshold fxn

Tobi-Alonso:
Speed-up of Multithreshold function using numpy vector operations.
Previous implementation of Multithreshold is now used to test the vectorized one using a random input and fixed thresholds.

Passing test inside of finn's docker:
- `python setup.py test --addopts "-k test_multi_thresholding.py"`

### DIFF
--- a/src/finn/custom_op/multithreshold.py
+++ b/src/finn/custom_op/multithreshold.py
@@ -33,16 +33,6 @@ from finn.core.datatype import DataType
 from finn.custom_op import CustomOp
 
 
-def compare(x, y):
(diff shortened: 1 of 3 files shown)


maltanar:
Thanks Tobi! Recording the speed-up from the vectorized multithreshold op here for reference:

```
Input tensor is (1, 256, 64, 64), threshold tensor is (1, 6)
Runtime non-vectorized:  14.160313844680786 s
Runtime vectorized:  0.03505516052246094 s
Speed-up:  403.94377414440396
```

PR #104, opened 2020-05-14
Changed how pyverilator is called, to support Vivado 2019.2

### LINKED ISSUE #89: rtlsim fails when using Vivado 2019.2

quetric:
This enables rtlsim with Vivado 2019.2 and fixes #89 

@maltanar I checked against several of the tests for dataflow ops (fclayer, maxpool, slidingwindow) and these pass. However, any block design TCLs which had been exported from Vivado 2019.1 may not work in 2019.2 and will need to be re-exported or modified. Any such cases should be picked up by deployment tests and I can fix the problematic TCLs if you point me to them.

### DIFF
--- a/docker/Dockerfile.finn_ci
+++ b/docker/Dockerfile.finn_ci
@@ -73,6 +73,8 @@ RUN mkdir -p $VIVADO_IP_CACHE
 WORKDIR /workspace/finn
 
 COPY finn_entrypoint.sh /usr/local/bin/
+COPY quicktest.sh /usr/local/bin/
 RUN chmod 755 /usr/local/bin/finn_entrypoint.sh
+RUN chmod 755 /usr/local/bin/quicktest.sh
 ENTRYPOINT ["finn_entrypoint.sh"]
 CMD ["bash"]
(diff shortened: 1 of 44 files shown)


maltanar:
I've made some minor changes (HLSCustomOp methods to get list of verilog files and paths, support for multiple files for IP stitched rtlsim) and was able to get rtlsim for test_end2end_tfc_w1a1_verify* tests using Vivado 2019.2. However, as expected PYNQ shell related transformations fail, since the .tcl scripts are all Vivado 2019.1.

I've verified that the PR doesn't break the end2end flows using 2019.1 so merging this in to start making 2019.2 support possible. There are more issues to be sorted (relating to PYNQ shell and possibly resource annotation passes) before FINN "officially" moves 2019.2 though.

PR #107, opened 2020-05-14
Implemented LabelSelect (TopK) HLS Op

### LINKED ISSUE #74: Add HLSCustomOp for LabelSelect_Batch

quetric:
Fixes #74 

Added a utility function to test.py as it will probably be useful in multiple tests.


maltanar:
Thanks Lucian. I've added assertions into the op so that those known-failing combinations are not used, and opened issue #116 to track this.

PR #110, opened 2020-05-15
transform that moves scalar linear operations past eltwise add

quetric:
… eltwise add


maltanar:
Thanks! Added an extra check in MoveScalarLinearPastEltwiseAdd to make sure the parameter tensors are scalars (the Add/Mul op itself does not guarantee this) and skip otherwise. This is probably too conservative, should be possible to extend to 1D mul/adds as needed.

PR #111, opened 2020-05-15
Added Addstreams HLS Op

quetric:
Required for elementwise addition


src/finn/custom_op/fpgadataflow/__init__.py
```diff
@@ -336,7 +336,7 @@ def toggle_clk(self, sim):
         sim.io.ap_clk = 1
         sim.io.ap_clk = 0
 
-    def rtlsim(self, sim, inp):
+    def rtlsim(self, sim, inp, inp2=None):
```
maltanar, line 339: once we have rtlsim infra that supports multiple inputs/outputs this should be updated, keeping it as is for now

PR #113, opened 2020-05-18
Feature/linear reorder for residual

Tobi-Alonso:
Depends on : #112  (first 2 commits)

Add checks in reorder transforms so they only apply changes in linear segments of graphs.
The same checks is used in all the transforms. Check that:
- neither of the nodes to swap were merging to branches
- there was no fork between those nodes

For this the helper functions is_fork_node and is_join_node were added to ModelWrapper.

Test cases added.

### DIFF
--- a/src/finn/core/modelwrapper.py
+++ b/src/finn/core/modelwrapper.py
@@ -328,6 +328,22 @@ class ModelWrapper:
         else:
             return None
 
+    def is_fork_node(self, node):
+        """Checks if the given node is a fork, that is, the node has multiple
+        direct successors"""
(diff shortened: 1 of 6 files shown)


maltanar:
Thanks Tobi, looks good. One thing that came to mind reviewing this was how the `is_join_node` is currently implemented. In ONNX there are two ways to specify constants, one with initializers (which FINN adopts) and one using `Constant` nodes to actually produce the constant. The current way of checking the direct predecessors will work when using the initializer method we adopt, but it will fail for any graphs that use `Constant` nodes. I'll open a separate issue about checking for `Constant` nodes.

PR #118, opened 2020-05-19
Feature/dupstreams batch

quetric:
Implements a HLS op for duplicating streams, useful for nonlinear topologies.

Yaman, with the addition of rtlsim_multi_io, I could modify the addstreams_batch code to use that and revert rtlsim to its original form (single input). Let me know what you prefer.


maltanar:
Thanks Lucian, I've moved the rtlsim_multi_io int utils. I'll merge this as-is for now and open a separate issue to migrate all HLSCustomOps to use rtlsim_multi_io so that we can get rid of the other variants.

PR #130, opened 2020-05-22
Feature/sort graph

Tobi-Alonso:
## Solves
Adds Transform to sort graphs topologically 

## Dependencies
- new package dep:  toposort  (tested with version 1.5)

## Test
transformation/test_sort_graph.py
  - if run as main does a performance test 

## Notes
- Small change to find_producer that should make it run twice as fast on average

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,5 @@ pre-commit
 pyverilator
 scipy
 sphinx
+toposort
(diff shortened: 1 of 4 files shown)


Tobi-Alonso:
Tests Fail because of new package dependency: toposort

Issue #123, opened 2020-05-20
New transformation: maxpool past multithreshols


Tobi-Alonso:
Maxpool can be moved past a multithreshold. As the bit width after a Multithreshold would be smaller, maxpool comparisons would require less hardware and potentially reduce the critical path of maxpool. 

Also, often, before a maxpool, there is a conv. by moving the maxpool past the multithreshold, conv and MT can be implemented with the mvau

maltanar:
This is provided by MoveMaxPoolPastMultiThreshold in https://github.com/Xilinx/finn/blob/dev/src/finn/transformation/streamline/reorder.py

PR #131, opened 2020-05-22
Feature/move max pool past multi threshold

Tobi-Alonso:
## Solves
Issue #123 
## Dependencies
PR #127 

## Test
  - transformation/ test_move_maxpool_past_multithreshold.py

## Notes
I think there may be problems if the output of the multithreshold is signed and the maxpool performs zero padding.
For this reason I'm checking the multithreshold attribute "out_dtype" to see if it is "UINTX".
May be out_bias should be check instead

onnx exection doesn't show there is a problem. I enter a matrix full of "-1" to a maxpool layer with padding and using onnxruntime to execute the node returns a matrix full of "-1". I would've expected zeros on the boarders

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ MANIFEST
 
 # Jenkins cfg dir
 /docker/jenkins_home
+
+# SSH key dir mounted into Docker
+/ssh_keys/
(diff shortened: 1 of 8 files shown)


maltanar:
Thanks Tobi. (MaxPool, MultiThreshold) reordering is an important transform and there are some tricky cases so I made a few changes here:

* use our built-in methods to check for signed/unsigned activation (DataType.signed())
* check for presence of padding as well as signed act before skipping. we do want reordering to happen for e.g. MaxPool with no padding -> bipolar MT
* check for non-decreasing thresholds and non-negative out_scale in MT, otherwise reordering won't work

PR #137, opened 2020-05-28
Feature/update to latest finn hlslib

auphelia:
Update finn to latest finn-hlslib commit version by adding resource type to ConvolutionInputGenerator and changing the format of saving weights for streaming MVAU (FCLayer in decoupled mode).


maltanar:
Let's keep the possibility to add directives by overriding the HLSCustomOp member fxn `ipgen_extra_directives`, we may need this for other cases in the future, even though it is not needed for the SWG memory resources anymore. I'll bring that back.

PR #138, opened 2020-06-03
Feature/convolution input generator kernel stride

Tobi-Alonso:
In order to add support for sliding window generator with kernel_size % stride != 0, `ConvolutionInputGenerator` make `ConvolutionInputGenerator_kernel_stride` HLS function call in those cases only (kernel_size % stride != 0) as this function is not as optimal as `ConvolutionInputGenerator`.

Remove verification for k % stride == 0 in several places.

**Problem:**
cppsim (finn) passes and `ConvolutionInputGenerator_kernel_stride` passes Vivado HLS cosim (hlslib test with same parameters) and has been implemented in hardware proving correct behavior. However, rtlsim in finn does not pass. 

Ideas of how to solve this:
- modify working HLS function: probably, It will be tricky to fix this and at the same time get the same performance and flexibility the current transform achieves
-  This is just a workaround: modify ` execute_node` of ` ConvolutionInputGenerator`  HLSCustomOp to execute cppsim for kernel_size%stride !=0 even if self.get_nodeattr("exec_mode") is "rtlsim". A warning could also be added to notify this situation
- use Vivado HLS cosim for these cases. Slower but it may be worth the effort to have HLS cosim as a backup as a simulator

### DIFF
--- a/src/finn/custom_op/fpgadataflow/convolutioninputgenerator.py
+++ b/src/finn/custom_op/fpgadataflow/convolutioninputgenerator.py
@@ -106,7 +106,6 @@ class ConvolutionInputGenerator(HLSCustomOp):
         pad = 0
         ofm_dim = compute_conv_output_dim(ifm_dim, k, stride, pad)
         assert ifm_ch % simd == 0, "SIMD must divide IFMChannels"
-        assert k % stride == 0, "stride must divide kernel size k"
         wf = int((k * k * ifm_ch) // simd)
         folded_oshape = (1, ofm_dim, ofm_dim, wf, simd)
         return folded_oshape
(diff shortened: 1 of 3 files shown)


Tobi-Alonso:
The test is fpgadataflow/test_fpgadataflow_convinputgenerator.py

Tobi-Alonso:
This [PR](https://github.com/Xilinx/finn-hlslib/pull/23) in finn-hlslib fixes the problem.
The following subset of the tests in fpgadataflow/test_fpgadataflow_convinputgenerator.py were previously failing and now are passing.

```python
@pytest.mark.parametrize("idt", [ DataType.INT2])
# kernel size
@pytest.mark.parametrize("k", [3,])
# input dimension
@pytest.mark.parametrize("ifm_dim", [6, 8])
# input channels
@pytest.mark.parametrize("ifm_ch", [ 4])  # , 2, 3, 4])
# Stride
@pytest.mark.parametrize("stride", [ 2])
# execution mode
@pytest.mark.parametrize("exec_mode", [ "rtlsim"])
# input channel parallelism ("SIMD")
@pytest.mark.parametrize("simd", [ 2,1])
```

PR #142, opened 2020-06-08
Feature/add channelwise op

Tobi-Alonso:
**Summary:** Add support for hlslib ChannelWiseOperation activation that allow us to have standalone functions with one parameter per channel

**Notes:**
- updated HLSLIB commit to last commit (we need the last one to update FMPadding_Batch)
- most of this is Lucian's. I adapted it to ChannelWiseOperation (originally coded for old ThresholdsActivation)


src/finn/transformation/fpgadataflow/convert_to_hls_layers.py
```diff
@@ -489,3 +491,113 @@ def apply(self, model):
...
+                    if idt.signed():
+                        odt = DataType.get_smallest_possible(2 * idt.min())
+                    else:
+                        odt = DataType.get_smallest_possible(2 * idt.max())
+                elif node.op_type == "Mul":
+                    func = "mul"
+                    if idt.signed():
+                        odt = DataType.get_smallest_possible(abs(idt.min()) * idt.min())
+                    else:
+                        odt = DataType.get_smallest_possible(idt.max() * idt.max())
```
maltanar, line 578: shouldn't you be also checking the data type for the parameter (ll_cinit's datatype) when deciding the output datatype? might make sense to skip any add/muls that don't have integer params too.

maltanar:
@quetric @Tobi-Alonso have you had a chance to look at my comment above? checking since other PRs depend on this one. I don't mind merging this as-is either but I prefer to be a little more conservative when converting things to HLS.

src/finn/transformation/fpgadataflow/convert_to_hls_layers.py
quetric: Yes this code assumes the parameter has the same datatype as the input, which was the case when I initially wrote it, but not true any more

Tobi-Alonso:
It would be great to have continuous bit-width data types, instead of just power of 2.
To support fc bias I had to "clip" that data type

maltanar:
I opened an issue to track the arbitrary bitwidth INT DataType issue here: 
https://github.com/Xilinx/finn/issues/179

maltanar:
The generated op test cases were failing because they were missing the paramDataType attribute. Please make sure the tests you added actually pass for the PR. I fixed this one myself to get faster progress on this PR since others depend on it.

Tobi-Alonso:
Sorry about that, totally forgot that test was there. I was using test_convert_to_hls_channelwise_layer.py  to test this as it covers the whole flow.

PR #143, opened 2020-06-08
Feature/fix globalAccpool batch shape

Tobi-Alonso:
**Depends on:**
- PR #142 (I branched from there to avoid possible conflicts with transformation/fpgadataflow/convert_to_hls_layers.py)

**Summary:** 
- Fix how was computed the output shape of GlobalAccPool_Batch
- Add InferGlobalAccPoolLayer transform to support the implementation of GlobalAveragePool

**Note:**
- This is Lucian's code


src/finn/custom_op/fpgadataflow/globalaccpool_batch.py
```diff
@@ -75,16 +75,19 @@ def get_folded_input_shape(self):
     def get_normal_output_shape(self):
         ch = self.get_nodeattr("NumChannels")
         vecs = list(self.get_nodeattr("numInputVectors"))
-        oshape = tuple([vecs[0]] + [ch])
+        if len(vecs) == 1:
+            oshape = tuple(vecs + [ch])
```
maltanar, line 79: is this a valid use case? in which cases(s) does this op get an op that looks like (1, ifm_ch)?

maltanar:
The conversion transformation introduced in this PR does not have a test, but since we've moved away from using non-quantized AvgPool as part of our example networks I'm fine with this for now. I made a comment above regarding the computation of the output shape when the input shape is missing the spatial dimension, but that is mostly for clarification and we can handle that later if it comes up

PR #144, opened 2020-06-10
Feature/onnx export quant avg pool

auphelia:
Add new CustomOp that corresponds to Brevitas QuantAvgPool2d and testing export and execution of that node.


maltanar:
@auphelia can you have a look at the failed tests here? One of them is the newly-added QuantAvgPool2d export (unsure if this is related to the already-known failures that you discussed with Alessandro already), but the other one occurs during test_end2end_cnv_w1a1. The second one may be related to the sanitize_quant_tensor functionality, or the change in behavior in InferDataTypes in this branch.

PR #146, opened 2020-06-15
Added custom hls op for DMA node, and transform to insert it

quetric:
Added a custom hls op what implements DMA functionality from the hlslib dma.h:

- input dma
- output dma
- wrapping dma, required for e.g. weight streaming from external memory

Note: simulation is not implemented for this node as it's unclear how to fit it within the stream in / stream out simulation harness. Open to suggestions on this. Also, there are no unit tests in this PR but a test will be submitted for multiple aspects of Vitis integration later on.

### DIFF
--- a/src/finn/custom_op/fpgadataflow/iodma.py
+++ b/src/finn/custom_op/fpgadataflow/iodma.py
@@ -1,7 +1,73 @@
+# Copyright (c) 2020, Xilinx
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+#
+# * Neither the name of FINN nor the names of its
+#   contributors may be used to endorse or promote products derived from
+#   this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
 import numpy as np
+import math
 from onnx import TensorProto, helper
 from finn.core.datatype import DataType
 from finn.custom_op.fpgadataflow import HLSCustomOp
+
+
+# the IODMA inerfaces a memory-mapped AXI interface and an AXI stream
+# direction "in": pulls data from AXI-MM to AXI stream
+# direction "out": pushes data from AXI stream to AXI-MM
+
+# DMA Addressing
+# - burst mode can be "wrap" or "increment"
+# - "increment" bursts will increment the address when moving to the next image
+# - "wrap" bursts will reinitialize the address to the start address,
+#   and are useful for e.g. streaming weights, where the same buffer is
+#   repeatedly read into the FPGA
+# - no additional alignment restrictions beyond anything specified in the AXI spec
+
+# Interfaces
+# - AXI-MM interface width (in bits) is specified by intfWidth
+# - AXI-Stream interface width (in bits) is specified by streamWidth
+# - If inftWidth and streamWidth are not equal, the DMA core performs
+#   width conversion by going up to the least common multiple of bitwidths
+#   e.g. intfWidth=32b -> 96b -> sreamWidth=24b
+# - transfers occur in multiples of the AXI-MM interface width, therefore
+#   the total number of bits in the tensor must be a multiple of intfWidth
+# - transfers occur in multiples of the AXI-Stream interface width, therefore
+#   the total number of bits in the tensor must be a multiple of streamWidth
+# - both interface widths must be a multiple of 8b (AXI protocol requirement)
+# - in most systems, intfWidth is also restricted to a power of 2 (e.g. Vitis)
+#   but this is not universal so we don't check here explicitly
+
+# Input/output tensor sizes shapes
+# - The data being moved is a tensor of shape numInputVectors+[NumChannels]
+# - The data type of the tensor elements is specified by dataType
+# - on the stream side
+#       -the normal shape is the same as the ONNX tensor attached to it
+#       -the folded shape is computed from the stream width and normal shape
+# - on the AXI-MM side
+#       -the normal shape is the same as the one on the stream side
+#       -the folded shape is not defined
 
 
 class IODMA(HLSCustomOp):
@@ -15,6 +81,8 @@ class IODMA(HLSCustomOp):
             "NumChannels": ("i", True, 0),
             # FINN input datatype
             "dataType": ("s", True, ""),
+            # Stream parameters
+            "streamWidth": ("i", False, 32),
             # DMA-specific parameters
             "intfWidth": ("i", False, 32),
             "burstMode": ("s", False, "increment"),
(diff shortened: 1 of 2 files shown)


src/finn/custom_op/fpgadataflow/iodma.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def get_nodeattr_types(self):
+        my_attrs = {
+            "NumChannels": ("i", True, 0),
+            # FINN input datatype
+            "dataType": ("s", True, ""),
+            # DMA-specific parameters
+            "intfWidth": ("i", False, 32),
+            "burstMode": ("s", False, "increment"),
+            "direction": ("s", False, "in"),
```
maltanar, line 21: add comments describing what these attributes do and which attribute value(s) are allowed

src/finn/custom_op/fpgadataflow/iodma.py
```diff
@@ -0,0 +1,189 @@
+import numpy as np
+from onnx import TensorProto, helper
+from finn.core.datatype import DataType
+from finn.custom_op.fpgadataflow import HLSCustomOp
+
+
```
maltanar, line 6: what are the assumptions about i/o shape, alignment etc for the DMA? any special considerations needed?

src/finn/custom_op/fpgadataflow/iodma.py
```diff
@@ -0,0 +1,189 @@
...
+        return self.get_normal_input_shape()
+
+    def get_folded_input_shape(self):
+        shape = list(self.get_normal_input_shape())
+        itype_bits = self.get_input_datatype().bitwidth()
+        intfw = self.get_nodeattr("intfWidth")
+        elems_per_word = intfw / itype_bits
+        fold_depth = round(shape[-1] / elems_per_word)
+        shape[-1] = fold_depth
+        shape.append(elems_per_word)
```
maltanar, line 44: I don't fully understand the logic here and L42 with the rounding (up? down?) is a bit alarming. what are the DMA alignment restrictions? what implications are there for packing the input/output arrays? do we assume that the innermost (packed) dimension is divisible by the intfWidth?

note that we have e.g. get_instream_width_padded() in the HLSCustomOp base class (returns input stream width padded to nearest 8 bits as required by AXI stream) which may be relevant here:
https://github.com/Xilinx/finn/blob/dev/src/finn/custom_op/fpgadataflow/__init__.py#L568

src/finn/custom_op/fpgadataflow/iodma.py
```diff
@@ -0,0 +1,189 @@
...
+        return tuple(shape)
+
+    def get_folded_output_shape(self):
+        return self.get_folded_input_shape()
```
maltanar, line 48: shouldn't the input and output folded shapes be different for the stream/mem DMA engines? or is it always guaranteed that stream width = intf width?

src/finn/custom_op/fpgadataflow/iodma.py
```diff
@@ -0,0 +1,189 @@
...
+    def get_outstream_width(self):
+        return self.get_instream_width()
+
+    def get_number_output_values(self):
+        oshape = self.get_normal_output_shape()
+        itype_bits = self.get_input_datatype().bitwidth()
+        intfw = self.get_nodeattr("intfWidth")
+        nelems = np.prod(oshape)
+        nbits = nelems * itype_bits
+        assert nbits % intfw == 0, "DMA: total transfer size must be word multiple"
```
maltanar, line 100: so from this line it looks like there is an alignment assumption/requirement s.t. the total transfer size is divisible by the interface width (btw np.prod on the packed output shape may be quicker for the num elems calculation here). 

1) how do we deal with cases where the first/last layers of a network may produce a nonaligned number of bits?
2) are there any extra considerations needed for AXI stream width? (internal stream widths always get padded to a multiple of 8, normally not a problem since the padded bits are just ignored in the next layer)

src/finn/custom_op/fpgadataflow/iodma.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def docompute(self):
+        direction = self.get_nodeattr("direction")
+        mode = self.get_nodeattr("burstMode")
+        if direction == "in":
+            if mode == "wrap":
+                func = "Mem2Stream_Batch_external_wmem"
+            else:
+                func = "Mem2Stream_Batch"
+        else:
```
maltanar, line 126: please restrict these to allowed predetermined values instead of using "else"

src/finn/custom_op/fpgadataflow/iodma.py
```diff
@@ -0,0 +1,189 @@
...
+            )
+        else:
+            self.code_gen_dict["$PRAGMAS$"].append(
+                "#pragma HLS INTERFACE axis port=in0"
+            )
+            self.code_gen_dict["$PRAGMAS$"].append(
+                "#pragma HLS INTERFACE m_axi offset=slave port=out"
+            )
+            self.code_gen_dict["$PRAGMAS$"].append(
+                "#pragma HLS INTERFACE s_axilite port=out bundle=control"
```
maltanar, line 173: suggest to make the AXI MM interface names controllable via attributes (default values can be the current ones here) for future proofing

src/finn/transformation/fpgadataflow/insert_iodma.py
```diff
@@ -0,0 +1,118 @@
...
+        final_node = model.find_producer(graph_out_name)
+        if final_node.op_type == "IODMA" and first_node.op_type == "IODMA":
+            # TODO maybe check the correctness of properties
+            return (model, False)
+        else:
+            if final_node.op_type != "IODMA":
+                out_shape = model.get_tensor_shape(graph_out_name)
+                out_dtype = model.get_tensor_datatype(graph_out_name)
+                # determine the feasible interface width
+                transfer_bits = np.prod(out_shape) * out_dtype.bitwidth()
```
maltanar, line 61: does the AXI stream padding need to be factored in here? 

see e.g
https://github.com/Xilinx/finn/blob/dev/src/finn/custom_op/fpgadataflow/__init__.py#L568
https://github.com/Xilinx/finn/blob/dev/src/finn/custom_op/fpgadataflow/__init__.py#L574

src/finn/transformation/fpgadataflow/insert_iodma.py
```diff
@@ -0,0 +1,118 @@
...
+                intfwidth = math.gcd(transfer_bits, self.max_intfwidth)
```
maltanar, line 62: what are the assumption(s) about the hardware shell here? can it accommodate arbitrary AXI MM widths as long as they are divisible by 8?

src/finn/transformation/fpgadataflow/insert_iodma.py
```diff
@@ -0,0 +1,118 @@
...
+                model.graph.value_info.append(final_node_out)
+                model.set_tensor_datatype(final_node_out.name, out_dtype)
+                # reroute final node output to final_node_out_name
+                final_node.output[0] = final_node_out.name
+                dma_node = oh.make_node(
+                    "IODMA",
+                    [final_node_out.name],
+                    [graph_out_name],
+                    numInputVectors=out_shape[:-1],
+                    NumChannels=out_shape[-1],
```
maltanar, line 79: may make sense to add a data layout check here (for NHWC or NC since we're assuming C is at -1) to make the assumptions clearer and future-proof

src/finn/transformation/fpgadataflow/insert_iodma.py
```diff
@@ -0,0 +1,118 @@
...
+                    direction="out",
+                    domain="finn",
+                    backend="fpgadataflow",
+                )
+                model.graph.node.append(dma_node)
+            if first_node.op_type != "IODMA":
+                in_shape = model.get_tensor_shape(graph_in_name)
+                in_dtype = model.get_tensor_datatype(graph_in_name)
+                # determine the feasible interface width
+                transfer_bits = np.prod(in_shape) * in_dtype.bitwidth()
```
maltanar, line 91: does the AXI stream padding need to be factored in here?

see e.g
https://github.com/Xilinx/finn/blob/dev/src/finn/custom_op/fpgadataflow/__init__.py#L568
https://github.com/Xilinx/finn/blob/dev/src/finn/custom_op/fpgadataflow/__init__.py#L574

src/finn/custom_op/fpgadataflow/iodma.py
quetric: Added feature documentation in comments, and the copyright header

src/finn/custom_op/fpgadataflow/iodma.py
quetric: documented features in comments before class declaration

src/finn/custom_op/fpgadataflow/iodma.py
quetric: I've documented some of the restrictions. The folded shapes were a bit tricky on the AXI-MM side since there is no guarantee that the AXI-MM interface width is an integer multiple of the element width. With this in mind, I added width conversion, so now the stream side has the same normal shape as the tensor it connects to, and the folded shape is defined by the stream width. The AXI-MM side has no folded shape (it cannot be defined here) but has the same normal shape as the tensor connected to the stream side.

src/finn/custom_op/fpgadataflow/iodma.py
quetric: Documented transfer size requirements with relation to intfWidth/streamWidth. This version of the hlslib dma (and therefore the IODMA node itself) won't work with streams widths which aren't a multiple of 8b and tensors whose total number of bits is not divisible by the AXI-MM interface width. This shouldn't be a problem for network inputs or outputs, but may be problematic for streaming weights later on. I would propose modifying the hlslib dma and the IODMA when use-cases require.

src/finn/transformation/fpgadataflow/insert_iodma.py
quetric: There is no restriction beyond the AXI one (multiples of 8b) and the fact that the transfer must be an integer number of words on the axi-mm interface (as the hlslib dma requires, and comments in iodma.py now indicate). Some shells (e.g. Vitis) require powers of 2 axi-mm widths, but this isn't general and I think it shouldn't be enforced here.

src/finn/custom_op/fpgadataflow/iodma.py
quetric: In fact folded shape can't reliably be defined for the axi-mm interface. Modified the code to define the folded shape on the stream side properly, since that is the side connected to any other FINN logic.

src/finn/transformation/fpgadataflow/insert_iodma.py
quetric: It would be more complicated than that, as some realignment logic would be required to remove the padding bits on the AXI-MM side. I suggest we leave this as is for now.

PR #147, opened 2020-06-15
Added features to tlastmarker and updated the insertion transform

quetric:
Added a number of features to the tlast marker:
- ability to use static or dynamic number of iterations. Dynamic is the default, and creates a axilite interface. static does not create an interface and uses the number of iterations specified in the attribute of the same name
- ability to use qdma_axis or ap_axiu for the stream interfaces. While they don't seem to differ, Vitis documentation makes a point of specifying that qdma_axis is to be used for streaming connections to the host, while ap_axiu is for kernel-to-kernel streaming
- ability to remove tlast instead of inserting it, if specified a "in" direction

These changes make the tlastmarker generally useful as a stream adapter for kernel to kernel/host communication in Vitis.

Depends on: #146 weakly, since the IODMA op name is checked for in the to_hls transform. This shouldnt cause any trouble though even if this PR is merged first

### DIFF
--- a/src/finn/custom_op/fpgadataflow/tlastmarker.py
+++ b/src/finn/custom_op/fpgadataflow/tlastmarker.py
@@ -51,8 +51,8 @@ class TLastMarker(HLSCustomOp):
             "StreamWidth": ("i", True, 0),
             # width of individual element in stream, in bits
             "ElemWidth": ("i", True, 0),
-            # Protocol: external or kernel2kernel
-            # Vitis docs recommend using qdma_axis for external, ap_axiu for kernel2kernel
+            # Protocol: external or internal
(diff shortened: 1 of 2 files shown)


maltanar:
Thanks Lucian. I'm merging this ahead of #146 since there was no hard constraint. I fixed a couple of issues myself, but for future PRs please:
*  reflect any changes you make to custom ops and/or transformations in the comments
*  use 1 and 0 instead of strings "true" and "false" for boolean attributes
*  specify the allowed values for properties (e.g. "internal" or "external") and check for unspecified/other values during codegen

Note that I modified the DynIters attribute to be 1/0 instead of "true"/"false", please update any dependencies accordingly.

PR #149, opened 2020-06-15
Added floorplanning transform

quetric:
Simple floorplanning transform that sets the partition_id attribute of dataflow nodes. For now, the transform will just ensure the design is Vitis-compatible: put each node that has a axilite interface in its own partition, such that no partition has more than one axilite. Each partition will subsequently be converted to a kernel during the Vitis build flow.

Depends on #146 and #147


maltanar:
Hey Lucian, the dependency on #146 is a weak one right? (I see the IODMA op is checked for in the transform but it's just a string comp so this will be fine even if the IODMA op does not exist).

We should add a basic test for this, as a side-test part of `test_fpgadataflow_ipstitch` tests for instance (so the floorplaning test would depend on the the checkpoint with HLS layers, it would add the DMA nodes, run the floorplanning transform and check that the partition numbers are as expected). The test result would not be used further as part of the end2end flow.

quetric:
The dependency on #146 was soft before, but I've added a test now which makes it a hard dependency. Since the main purpose of the floorplanning transform is to put DMAs into separate partitions from the other logic, there's no way to do a meaningful test without enforcing the dependency. In any case, this isn't super urgent.

maltanar:
I added the IODMA HLSCustomOp to the registry, but otherwise this looks all good. I'm happy with the rudimentary test in test_fpgadataflow_ip_stitch.py for now, once we have fancier partitioning logic that should probably have its own test, perhaps based on one of the end2end examples.

PR #156, opened 2020-06-18
[HLSCustomOp] Update FMPadding_Batch. New attribute: SIMD

Tobi-Alonso:
Update FMPadding_Batch to the last version of the HLSLIB function.
New:
-  SIMD: attribute to control Input parallelism
- Test new feature

PR #157, opened 2020-06-18
Add support for pooling layers with stride != kernel_size

Tobi-Alonso:
## Dependencies
- PR #156 

## New
- Add new `Pool_Batch` op to support pooling layers with kernel_size != stride
- Add InferPool_Batch transformation to infer Pool_Batch op

## Test
- New test: fpgadataflow/test_convert_to_hls_pool_batch
  - conversion of MaxPool with stride != kernel_size down to HLS ops

## Note:
- Remove print used for debug of prepare_cppsim


Tobi-Alonso:
test_brevitas_avg_pool_export[8-4-4-4-False-1-2] FAILED [ 30%]
The failing test has nothing to do with this PR.

maltanar:
> test_brevitas_avg_pool_export[8-4-4-4-False-1-2] FAILED [ 30%]
> The failing test has nothing to do with this PR.

Thanks Tobi, yes I'm aware. Giuseppe and Alessandro are working on debugging the QuantAvgPool2d export problems.

src/finn/custom_op/fpgadataflow/pool_batch.py
```diff
@@ -0,0 +1,395 @@
...
+    # * depthwise SWG: (1, OFMDim, OFMDim, IFMChannels/SIMD, K, K, SIMD)
+
+    Channels can be folded using PE (SIMD from the input perspective)
+    TODO: doc
+    """
+
+    def get_nodeattr_types(self):
+        my_attrs = {
+            "Channels": ("i", True, 0),
+            "PE": ("i", True, 1),
```
maltanar, line 57: this is a bit confusing -- why is it called PE here and SIMD in the comment above? why not just call this SIMD if it's input parallelism?

src/finn/custom_op/fpgadataflow/pool_batch.py
```diff
@@ -0,0 +1,395 @@
...
+            "KernelSize": ("i", True, 0),
+            # Function:
+            #  - MaxPool
+            #  - AvgPool (not yet supported, but HLSLIB does)
+            #  - AccPool (not yet supported, but HLSLIB does)
+            "Function": ("s", True, ""),
+            "OutImgDim": ("i", True, 0),
+            # FINN DataTypes for inputs/outputs
+            "dataType": ("s", True, ""),
```
maltanar, line 66: I believe the hlslib Pool_batch support different i/o datatypes? For maxpool idt=odt is fine but this won't always be the case (e.g. AvgPool, AccPool)

src/finn/custom_op/fpgadataflow/pool_batch.py
```diff
@@ -0,0 +1,395 @@
...
+                dims=values.shape,
+                vals=values.flatten().astype(float),
+            ),
+        )
+
+    def infer_node_datatype(self, model):
+        node = self.onnx_node
+        # data type stays the same
+        dtype = model.get_tensor_datatype(node.input[0])
+        model.set_tensor_datatype(node.output[0], dtype)
```
maltanar, line 163: does it? doesn't Pool_batch support different odt/idt types? there should be a check here for fxn=MaxPool at least

src/finn/custom_op/fpgadataflow/pool_batch.py
Tobi-Alonso: Yes, that's true. But I consider to compute the output datatype within the HLSCustomOp

src/finn/custom_op/fpgadataflow/pool_batch.py
Tobi-Alonso: that's and error. It should be:

```python
dtype = self.get_output_datatype()
model.set_tensor_datatype(node.output[0], dtype)
```

the check is on `get_output_datatype`

PR #158, opened 2020-06-18
Feature/move chw mul past dw conv

auphelia:
Add transformation to move channelwise multiplication past depthwise convolution node. Add test for new transformation.


src/finn/transformation/streamline/reorder.py
```diff
@@ -338,6 +338,60 @@ def apply(self, model):
...
+            ):
+                consumer = model.find_consumer(n.output[0])
+                if (
+                    consumer is not None
+                    and consumer.op_type == "Conv"
+                    and not model.is_join_node(consumer)
+                ):
+                    mul_weight_name = n.input[1]
+                    A = model.get_initializer(mul_weight_name)
+                    assert A is not None, "Initializer for mul weights is not set."
```
maltanar, line 364: suggest "if A is None, continue (e.g. skip to next loop iteration)" instead of assertion, e.g. when we find a pattern that we can't process we should simply go to the next node instead of raising an assertion error

src/finn/transformation/streamline/reorder.py
```diff
@@ -338,6 +338,60 @@ def apply(self, model):
...
+                    conv_node = consumer
+                    mul_node = n
+                    start_name = mul_node.input[0]
+                    conv_in_name = conv_node.input[0]
+                    conv_in_shape = model.get_tensor_shape(conv_in_name)
+                    ifm_ch = conv_in_shape[1]
+                    group_attribute = get_by_name(consumer.attribute, "group").i
```
maltanar, line 371: is the "group" attribute always specified? may want to do a None check here

src/finn/transformation/streamline/reorder.py
```diff
@@ -338,6 +338,60 @@ def apply(self, model):
...
+                    conv_out_name = conv_node.output[0]
+                    conv_out_shape = model.get_tensor_shape(conv_out_name)
+                    if np.prod(A.shape) == ifm_ch == group_attribute:
```
maltanar, line 374: suggest a stronger check on A.shape instead of looking at np.prod(A.shape); e.g. A.shape == (1, ifm_ch, 1, 1)

src/finn/transformation/streamline/reorder.py
```diff
@@ -338,6 +338,60 @@ def apply(self, model):
...
+                        # rewire mul input to be conv input
+                        conv_node.input[0] = start_name
+                        model.set_tensor_shape(start_name, conv_in_shape)
+                        # use old conv input tensor as conv output
+                        conv_node.output[0] = conv_in_name
+                        model.set_tensor_shape(conv_in_name, conv_out_shape)
+                        # use new conv output as new mul node input
+                        mul_node.input[0] = conv_in_name
+                        # use old conv output as new mul node output
+                        mul_node.output[0] = conv_out_name
```
maltanar, line 386: are the DataTypes for all tensors (start, intermediate, end) correctly set? start shouldn't really change but the end and intermediate ones?

(start) -> Mul -> (intermediate: dtype1) -> Conv -> (end: dtype2)
to
(start) -> Conv -> (intermediate) -> Mul -> (end)

if the post-reordering datatypes are tricky to determine it's better to set them to float32

PR #165, opened 2020-06-24
Feature/quant avg pool2d nhwc

auphelia:
Additional data layout for QuantAvgPool2d node and transformation to change data layout from (N, C, H, W) to (N, H, W, C) for this node (produces also surrounding transpose nodes).


src/finn/custom_op/quantavgpool2d.py
```diff
@@ -17,19 +18,48 @@ def get_nodeattr_types(self):
             "ibits": ("i", True, 1),
             "obits": ("i", True, 1),
             "signed": ("i", True, 0),
+            "data_layout": ("s", False, "NCHW"),
```
maltanar, line 21: add comments to attributes and specify what the allowed values are

maltanar: also, please add specific support for this node/attribute to:
https://github.com/Xilinx/finn/blob/f525398cbc3bb1a6b83be51b11e7923ed7c8870a/src/finn/transformation/infer_data_layouts.py#L41

src/finn/custom_op/quantavgpool2d.py
```diff
@@ -17,19 +18,48 @@ def get_nodeattr_types(self):
...
+                    name="const_tensor",
+                    data_type=TensorProto.FLOAT,
+                    dims=values.shape,
+                    vals=values.flatten().astype(float),
+                ),
+            )
+
+        else:
+            raise Exception(
+                """Datalayout for QuantAvgPool2d is set to an unvalid value.
```
maltanar, line 60: unvalid -> invalid

PR #178, opened 2020-07-01
Feature/merge models

auphelia:
Add transformation to merge two models. The model that gets passed as argument to the transformation is inserted before the model the transformation is applied to.


src/finn/transformation/merge_onnx_models.py
```diff
@@ -0,0 +1,171 @@
...
+    def apply(self, model):
+        graph_modified = False
+        pre_model = self.pre_model
+
+        # make model values unique to avoid any conflict
+        (pre_model, model) = _make_model_values_unique(pre_model, model)
+
+        # check if models can be merged
+        output_model_a = pre_model.graph.output[0].name
+        input_model_b = model.graph.input[0].name
```
maltanar, line 103: this will only connect 1 output of the pre graph to 1 input of the post graph. what if they have more outputs/inputs? this limitation should be specified in the transformation description.

you may also want to check num dynamic outputs in pre == num dynamic inputs in post. dynamic means that tensor has no initializer associated with it.

tests/transformation/test_merge_onnx_models.py
```diff
@@ -0,0 +1,97 @@
...
+        value_info=[a0, a1],
+    )
+
+    model2 = helper.make_model(graph, producer_name="model2")
+    model2 = ModelWrapper(model2)
+    # initialize model2
+    a0_value = np.random.uniform(low=0.1, high=0.99, size=(1)).astype(np.float32)
+    model2.set_initializer("a0", a0_value)
+    a1_value = 1.0 / a0_value
+    model2.set_initializer("a1", a1_value)
```
maltanar, line 73: due to float rounding there is no guarantee that x * A -> x / (A) will give x

tests/transformation/test_merge_onnx_models.py
```diff
@@ -0,0 +1,97 @@
...
+from finn.core.modelwrapper import ModelWrapper
+from finn.transformation.infer_shapes import InferShapes
+from finn.transformation.general import GiveReadableTensorNames, GiveUniqueNodeNames
+from finn.transformation.merge_onnx_models import MergeONNXModels
+import finn.core.onnx_exec as oxe
+
+
+def test_merge_onnx_models():
+    # load first model
+    raw_m = get_data("finn", "data/onnx/mnist-conv/model.onnx")
```
maltanar, line 43: suggest to use a test that exercises more FINN-specific things like quantization and sparsity annotations and checks that the merged model still contains those.

tests/transformation/test_merge_onnx_models.py
```diff
@@ -0,0 +1,97 @@
...
+    a0_value = np.random.uniform(low=0.1, high=0.99, size=(1)).astype(np.float32)
+    model2.set_initializer("a0", a0_value)
+    a1_value = 1.0 / a0_value
+    model2.set_initializer("a1", a1_value)
+    model2 = model2.transform(InferShapes())
+    model2 = model2.transform(GiveUniqueNodeNames())
+    model2 = model2.transform(GiveReadableTensorNames())
+
+    # simulate the models before the merging and pass the output of model2 to model1
+    inp_values = np.random.uniform(low=-1, high=1, size=tuple(shape)).astype(np.float32)
```
maltanar, line 79: when using real NN models as part of tests, it's better to use real test data instead of synthetic/random data. NNs are nonlinear in a very specific way and using random data may not be able to uncover if the NN is actually broken.

src/finn/transformation/merge_onnx_models.py
```diff
@@ -0,0 +1,171 @@
...
+from finn.transformation.infer_shapes import InferShapes
+from finn.transformation.infer_datatypes import InferDataTypes
+from finn.transformation.general import (
+    GiveReadableTensorNames,
+    GiveUniqueNodeNames,
+    GiveUniqueParameterTensors,
+)
+
+
+def _make_model_values_unique(model1, model2):
```
maltanar, line 44: my suggestion is to simply randomize all the tensor names in deep copies of both models, ignore any node name overlaps (there is no need for them to be unique -- indeed they can all be empty) and remove this function. this should make it easier to recover quantization/sparsity annotations.

Issue #116, opened 2020-05-18
Some LabelSelect configs produce incorrect results


maltanar:
The `LabelSelect_Batch` HLSCustomOp seems to be working fine for unsigned inputs and PE=1, but some of the tests fail when going to other configs. This may be related to the underlying hlslib op, how input/output is interpreted in FINN, or a combination of both.

Example failing testcase:

```
test_fpgadataflow_labelselect with the following config:
idt = <DataType.INT16: 14>, labels = 10, fold = -1, k = 5, exec_mode = 'rtlsim'

(Pdb) x
array([[-12043.,  30370., -23559., -30747., -11173.,  30686.,  19723.,
         -8171., -16169.,    809.]], dtype=float32)
(Pdb) y
array([[5., 1., 6., 9., 0.]], dtype=float32)

```

Here the first for of the top-5 inputs are correct, but the last one corresponding to the negative value in the top-5 list is produced to be index 0.

PR #182, opened 2020-07-02
Feature/fix Labelselect folding and optimize output

Tobi-Alonso:
Solves: #116 

- fix input folding != complete folding.
- Remove restriction on signed DataTypes as is the problem is fixed in hlslib. 
- Add Variable output width feature as hlslib now supports it

## Test
- Add signed input and not completely folded input test cases


maltanar:
Thanks Tobi. I changed one of the test options from 1000 to 100 labels since this creates rtlsim test cases that take very long to run. Unit tests shouldn't take more than a few minutes each.

PR #185, opened 2020-07-03
Modified docker-related scripts to enable Vitis in the container

quetric:
Added optional paths to Vitis installation and DSA installation. These are required if users want to build with Vitis.


maltanar:
Thanks Lucian. The error messages that appear when starting the container without Vitis may worry the user so I modified the scripts so that Vitis/Vivado are only mounted if the env.vars are specified.

PR #188, opened 2020-07-03
Updated stitching transform

quetric:
The Stitch IP transform was modified to support:

- Vitis IP generation
- IPs with multiple inputs, required for external weights

These changes require the rtls interface name retrieval functionality so **depends on #186 and #187**. Because input/output names can be arbitrary in the general sense, I chose to name the stream inputs and outputs of the stitched IP as [m|s]_axi_<idx> where idx starts from 0. For a linear graph with no external weights, this will result in a stitched ip with m_axis_0 and s_axis_0. Also renamed clock and reset ports to ap_clk and ap_rst_n, required for Vitis compatibility.

Also updated other transforms as a result of changing port names.

Tested up to and including rtlsim on the stitched IP, but pynq project generation and synthesis don't work for me as I dont have vivado 2019.1


maltanar:
This broke the MakePYNQProject transformation because the interface names there weren't updated, now fixed.

maltanar:
To make things a bit more future-proof I added a metadata_prop at the top-level stitched ONNX returned from the `CreateStitchedIP` transformation that lists the top-level interface names:

```
{'clk': ['ap_clk'], 'rst': ['ap_rst_n'], 's_axis': ['s_axis_0'], 'm_axis': ['m_axis_0'], 'aximm': [], 'axilite': ['s_axi_control']}
```

I'll modify the MakePYNQProject transform to make use of this `metadata_prop` instead of using fixed args then merge.

PR #193, opened 2020-07-07
[Docker] infer vivado path from path if present

Tobi-Alonso:
infer vivado path from path if present


maltanar:
Let's keep the VIVADO_PATH setting as a manual requirement. There are cases where an autodetected Vivado on PATH may cause issues (e.g. older/unsupported version, preference for difference version inside container), and the user may prefer to not have vivado at all inside the container even though it exists on the host.

PR #197, opened 2020-07-07
[Transformation] New transform to move linear ops past several ops

Tobi-Alonso:
New transform `MoveScalarLinearPastInvariants` to move linear ops past "GlobalAveragePool" ,"Reshape" ,"Transpose" and "Flatten"

## Notes
- Lucian's code
- It was needed when our use case used "GlobalAveragePool" instead of "QuantAvPool". Non needed for this case anymore.

### DIFF
--- a/src/finn/transformation/streamline/reorder.py
+++ b/src/finn/transformation/streamline/reorder.py
@@ -560,7 +560,9 @@ class MoveScalarLinearPastInvariants(Transformation):
                     graph_modified = True
                 else:
                     continue
-        model = model.transform(InferShapes())
+        if graph_modified:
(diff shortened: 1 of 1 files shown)


maltanar:
Added calls to `InferDataTypes` and `InferShapes` before exiting transform if graph was modified. Note that this may still lead to incorrect `DataType` annotations around the transformed tensors, e.g. if a scalar `Mul` originally had a `INT4` output annotated this transform will take that away.

maltanar:
Hopefully the output datatype issue is solved with this last commit. This should be checked in a future test, but since this PR is a dependency for other important PRs for upcoming work like #200 I am merging it as-is.

PR #198, opened 2020-07-07
Feature/infer nchw multi threshold

Tobi-Alonso:
Depends on: #192 (to avoid conflicts)
infer `Thresholding_Batch` inserting transposes if `MultiThreshold` in/out data type is NCHW


maltanar:
The failed check on this was due to a server/download failure, I've checked it locally and it's all good.

PR #199, opened 2020-07-07
Added transform for building with custom (non-Pynq) Zynq shell

quetric:
Adds capability to build for Zynq using a custom shell, using IODMAs similar to how Vitis builds work.

Depends on PRs up to #188.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ MANIFEST
 
 # SSH key dir mounted into Docker
 /ssh_keys/
+
+# PYNQ board files
+/board_files/
--- a/docker/Dockerfile.finn_ci
+++ b/docker/Dockerfile.finn_ci
@@ -37,7 +37,7 @@ RUN apt-get update
 RUN apt-get -y upgrade
 RUN apt-get install -y build-essential libglib2.0-0 libsm6 libxext6 libxrender-dev
 RUN apt-get install -y verilator zsh
-RUN apt-get -y install sshpass
+RUN apt-get -y install sshpass wget unzip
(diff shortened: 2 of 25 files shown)


maltanar:
The failed PR CI run above is unrelated to the contents of the PR, just an intermittent server failure.

maltanar:
@quetric after some fixes (see above) I was able to build bitfiles for Pynq-Z1 but interfacing them with the driver took some manual effort. The primary issue was the idma/odma units (created from the first and last StreamingDataflowPartition with one IODMA and one TLastMarker) -- these show up as `DefaultIP` under PYNQ, and since they have no C function signature they were not directly callable using the `.start()` method. Did this somehow work out of the box in Alveo, or is the DMA synthesis handled differently for Alveo?

I had to manually manipulate the control regs like this to get it working:
```
            self.idma.write(0x10, self.ibuf_packed_device.device_address)
            self.idma.write(0x1c, self.N)
            self.odma.write(0x10, self.obuf_packed_device.device_address)
            self.odma.write(0x1c, self.N)
            self.idma.write(0x00, 1)
            self.odma.write(0x00, 1)
            while self.odma.read(0x00) and 1 == 0:
                pass
```

quetric:
On alveo a function signature is generated during XO packaging which enables pynq to understand how to call the DMAs. I don't know what the equivalent is for Zynq though.

maltanar:
ZynqBuild now has three end-to-end tests (TFC-w1a1, TFC-w2a2, CNV-w1a1) and I'm happy to report all the tests are passing.

PR #207, opened 2020-07-27
Feature/expected cycles per layer

auphelia:
Add analysis pass to return a dictionary with node as key and expected cycles for set folding as value. Add for each HLSCustomOp a function that returns the expected cycles, integrate testing into existing unit tests for the HLSCustomOps.

### DIFF
--- a/src/finn/analysis/fpgadataflow/exp_cycles_per_layer.py
+++ b/src/finn/analysis/fpgadataflow/exp_cycles_per_layer.py
@@ -1,0 +1,44 @@
+# Copyright (c) 2020, Xilinx
+# All rights reserved.
+#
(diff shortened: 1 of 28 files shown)


quetric:
Is it a requirement to use `str(node)` as a key in the output dictionary? The `node.name` would be much easier to parse and understand.

maltanar:
I agree, assuming you mean this:

`cycle_dict[str(node)] = inst.get_exp_cycles()`

This was put in place since nodes have no guaranteed unique name by default, whereas we could use the string representation of it as a hashable dict key. But we can guarantee that all nodes have unique names by calling `GiveUniqueNodeNames` -- I'll update this and add a comment.

PR #213, opened 2020-08-12
Updates to resource estimates

quetric:
Implemented a best-effort folder which sets PE and SIMD, and mem_mode/ram_style according to simple rules.


quetric:
@maltanar refactored PR. Depends on PR 7 from finn-base, the two must be merged together

PR #219, opened 2020-08-24
Add capability to specify floorplan manually through a JSON file

quetric:
Added an analysis pass which generates a JSON file with all layer names and floorplanning parameters. This can be edited by the user and passed to to the Floorplan transform.

Depends on PR 5 in finn-base


maltanar:
I'm going to take this for a test drive using the Vitis nightly cnv-w1a1 as an example.

quetric:
@maltanar I've restructured the PR - it no longer depends on #213 but depends on a change to finn-base which i've submitted separately (PR 5)

PR #226, opened 2020-09-04
Feature/insertfifo nonlinear

Tobi-Alonso:
Add support for non linear graphs in insert_fifo transf


quetric:
It seems the features proposed here got into dev through other PRs, so closing.

PR #228, opened 2020-09-04
Add optional Vitis debug infrastructure

### LINKED ISSUE #222: Add debug flag to VitisBuild

quetric:
Fixes #222

### DIFF
--- a/src/finn/transformation/fpgadataflow/vitis_build.py
+++ b/src/finn/transformation/fpgadataflow/vitis_build.py
@@ -177,11 +177,15 @@ class VitisLink(Transformation):
     ModelProto's metadata_props field with the XCLBIN full path as value.
     """
 
-    def __init__(self, platform, f_mhz=200, strategy=VitisOptStrategy.PERFORMANCE):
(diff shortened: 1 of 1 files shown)


src/finn/transformation/fpgadataflow/vitis_build.py
```diff
@@ -267,12 +276,13 @@ def apply(self, model):
...
                 "v++ -t hw --platform %s --link %s"
                 " --kernel_frequency %d --config config.txt --optimize %s"
-                " --save-temps -R2\n"
+                " --save-temps -R2 %s\n"
                 % (
                     self.platform,
                     " ".join(object_files),
                     self.f_mhz,
                     self.strategy.value,
+                    debug_commands,
```
maltanar, line 285: shouldn't this be something like `" ".join(debug_commands)`? otherwise we get a command line call that uses Python list syntax.

quetric: Yes that's correct, I'll fix

PR #230, opened 2020-09-17
Memstream ip fix

jterry-x:
Added design files for use in simulation as well.  Vivado behavioral sim is otherwise unable to find the memstream IP during compilation.

Note this was done using version 2020.1 while previous ip was packaged with 2019.1.3


maltanar:
Thank you John! I did some minor fixes for the merge (e.g. there were a few user-specific paths in the component.xml) but this variant seems to work for both synthesis and simulation.

PR #232, opened 2020-09-21
Automatically discover the correct depth of FIFOs in the dataflow

quetric:
Uses RTLsim to discover what are the minimum depths of FIFOs between layers. The methodology is: 

1. deepcopy model, save original model
2. insert stream infrastructure (DWCs/FIFOs) 
3. set all FIFOs to very deep - 16K entries
4. run stitched RTLsim until pipeline is full, keeping track of maximum occupancy for each FIFO
5. annotate the original model with in/outFIFODepth parameters corresponding to depths discovered during rtlsim

Tests pass for all BNNs but I have not deployed these to hardware to compare the performance to expected values.

This PR depends on merging PR #2 in maltanar/pyverilator, which enables reading internal design signals during RTLsim


maltanar:
Thanks Lucian. I've pushed some more commits to get this working with the latest dev+finn-base separation, and was able to get the TFC-w1a1 example running.

However, I get the following for CNV-w1a1 by running `python setup.py test --addopts "-k 'End2End and cnv-1-1' --pdb"`:

```
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_export[cnv-1-1] PASSED [  4%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_import_and_tidy[cnv-1-1] PASSED [  8%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_add_pre_and_postproc[cnv-1-1] PASSED [ 12%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_streamline[cnv-1-1] PASSED [ 16%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_convert_to_hls_layers[cnv-1-1] PASSED [ 20%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_create_dataflow_partition[cnv-1-1] PASSED [ 25%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_fold[cnv-1-1] PASSED [ 29%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_cppsim[cnv-1-1] PASSED [ 33%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_ipgen[zynq-cnv-1-1] PASSED [ 37%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_ipgen[alveo-cnv-1-1] SKIPPED [ 41%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_set_fifo_depths[zynq-cnv-1-1] PASSED [ 45%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_set_fifo_depths[alveo-cnv-1-1] SKIPPED [ 50%]
tests/end2end/test_end2end_bnn_pynq.py::TestEnd2End::test_ipstitch_rtlsim[zynq-cnv-1-1] FAILED [ 54%]
```

The IP-stitched rtlsim is failing with:

```
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464: Cannot find include file: axis_infrastructure_v1_1_0.vh
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464: Looked in:
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/axis_infrastructure_v1_1_0.vh
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/axis_infrastructure_v1_1_0.vh.v
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/axis_infrastructure_v1_1_0.vh.sv
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       axis_infrastructure_v1_1_0.vh
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       axis_infrastructure_v1_1_0.vh.v
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       axis_infrastructure_v1_1_0.vh.sv
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       /tmp/finn_dev_maltanar/pyverilator_ipstitched_39hjbk49/axis_infrastructure_v1_1_0.vh
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       /tmp/finn_dev_maltanar/pyverilator_ipstitched_39hjbk49/axis_infrastructure_v1_1_0.vh.v
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:50464:       /tmp/finn_dev_maltanar/pyverilator_ipstitched_39hjbk49/axis_infrastructure_v1_1_0.vh.sv
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:90942: Cannot find include file: axis_infrastructure_v1_1_0.vh
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:91139: Cannot find include file: axis_infrastructure_v1_1_0.vh
%Error: /tmp/finn_dev_maltanar/vivado_stitch_proj_txmsgr0u/finn_design_wrapper.v:91321: Cannot find include file: axis_infrastructure_v1_1_0.vh
```

Any idea why? The FIFO sizing works and it also needs to do stitched rtlsim so not sure if something is called differently.

quetric:
Hi Yaman. This is because the design is configured with FIFOs from Vivado IP library after sizing. There is a parameter for the maximum admitted depth implementable with QSRL FIFO, by default 256. I assume the failures are for CNV which needs deeper fifos.

I see multiple ways of solving this: 
* look for Vivado IP HLS sources for rtlsim - i'm not sure these are available though
* set the threshold very high in sizing, but the resulting implementation is suboptimal
* set the fifo implementation_mode from vivado to rtl at the beginning of the rtlsim transform

I've implemented the third approach and can push code for it if you think it's appropriate.

maltanar:
OK, that was indeed the issue. I went for a form of option 3 as well, just set all FIFO `impl_style` to `rtl` in the test that calls rtlsim. I also added an assertion to `prepare_rtlsim` into StreamingFIFO so that we catch this earlier and produce a more interpretable error message.

I was able to get CNV-w1a1 as well with these fixes. I'll run the rest of the BNN-PYNQ tests overnight. I noticed that some of the FIFOs end up being relatively deep (e.g. 1024), CNV-w2a2 may not fit onto Pynq-Z1 with those.

maltanar:
CNV-wxa2 variants are failing due to restrictions on Vivado FIFO IP depths:

`ERROR: [IP_Flow 19-3461] Value '5120' is out of the range for parameter 'FIFO depth(FIFO_DEPTH)' for BD Cell 'StreamingDataflowPartition_1_StreamingFIFO_3/fifo' . Valid values are - 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768`

This one seems tricky. I propose to modify the `optimize_depth` function so that it rounds the depth to the nearest of the values above. Also, a FIFO depth of 5120 for cnv-w2a2 seems excessive, I wonder if we should be doing the FIFO sizing algorithm differently, such as monitoring the average (instead of max) occupancy.

quetric:
I agree with your approach, i'll modify and push. Do you know where the large FIFO is in W2A2? If it is the first FIFO, the large depth may result result from pushing data into the dataflow slightly faster than the average processing speed. The excess will accumulate into the FIFOs before the bottleneck node.

I'm not sure about using average instead of max, for some things (e.g. bypass FIFOs for residual nets) you actually want the max to prevent deadlock.

maltanar:
Based on input from Giulio I've added an additional transformation `CapConvolutionFIFODepths` to limit the sizes of SWG- > MVAU FIFOs since the SWG already has some built in buffering. I've also split out `RemoveShallowFIFOs` as a separate transformation as it may come in handy for the future.

All the BNN-PYNQ nets including cnv-w2a2 now fit on Pynq-Z1 and we see the expected performance on hardware. The resulting HW footprint is quite similar too, so I actually removed all of the manual FIFO size settings from the end2end tests. 

Thanks for the contribution!

Ba1tu3han:
Hello All,

I have encountered a issue that may be related to this commit. 

I use v0.10-dev version of FINN and compile CNV network to target Pynq-Z2. I use custom configuration file and I am sure it fit to the HW ('total': {'BRAM_18K': 103.0, 'LUT': 45935.0, 'URAM': 0.0, 'DSP': 0.0}}). I have problem with W2A2 althought W1A1 compilied well.

Could you give me recommendation ? @maltanar 

image

PR #234, opened 2020-09-24
Enable and test writable weights in decoupled mode

quetric:
Implement proper write access to weights through the decoupled weights memstreamer config interface.


maltanar:
I was having a look at this branch to see if I could start making a pyverilator AXI lite read/write-based testbench for runtime-writable weights, but got some errors building the tfc-w1a1 example due to a missing file:

`/workspace/finn/finn-rtllib/memstream/gui/memstream_v1_0.gtcl`

It's referenced from `component.xml` and the xgui tcl script, but doesn't look like it's added to the repo?

quetric:
Added that file now. I've tested the streamer writes/reads through axilite in hardware on pynq-z2, no problems reported by AXI protocol checker and data is correctly written to the streamer. Marking ready for review.

PR #238, opened 2020-10-14
Refactor and test external weights support

quetric:


Tobi-Alonso:
There's one thing missing from the driver generation.
The driver checks the "normal folding" of the input.
By the time a model gets to Zynk/Vitisbuild (when driver generation is called), it has possibly lost non-fpga nodes
Example, tfc, it has a flatten. Then, ZynqBuild sees a model with a flattened input and this shape is coded in the driver.
To keep this input fold check, a possibility might be to add metadata to the model regarding the input shape in the early stages, but other problems can arise. For example, if there's a transpose node, but a similar work around is possible in this case.

Probably the best thing is having a higher level driver that handles non-fpga nodes, which calls the driver generated by MakePYNQDriver. (As far as I remember finn execution worked like that). Then, the higher level driver would detect an input with a different dim. order, for example.

maltanar:
Thanks @Tobi-Alonso and @quetric for this PR. It took some time to go through this but it's finally ready to be merged into `dev`. I've added the fixes to make this work and made a separate end-to-end test for this (see `test/end2end/test_ext_weights.py`) to verify the execution, using `tfc-w1a1` with 2 external layers. I've not been able to test this on Alveo yet but any needed fixes for that can be a separate PR.

PR #246, opened 2020-11-02
[VitisBuild] Change memtype for U50 from DDR to HBM

auphelia:
If the selected board is U50, change the memory banks to HBM instead of DDR.


maltanar:
Thanks for the PR Jakoba. Two comments:

1) Can you re-target this PR against `dev` instead of `master`?

2) It's best to determine if the platform is U50 in another way instead of checking the the environment variable for `ALVEO_BOARD=U50`. I would use a check like `"u50" in self.platform`. Even better would be to create a function or dictionary that returns "HBM" or "DDR" given the platform string.

maltanar:
#314 which just got merged should sort out these issues for U50 so I'm closing this PR. Please feel free to reopen if something is still missing.

PR #270, opened 2021-01-29
Added suggestions form testing

HenniOVP:
Hi,

here are the changes, which I would suggest after testing the three notebooks.

All the best,
Hendrik


maltanar:
Thanks Hendrik, much appreciated!

PR #280, opened 2021-02-11
Fix ZCU102 support in Vivado shell project

fpjentzsch:
The ZCU102 board is officially supported, but the board part setting was missing in the respective Tcl script template.

### DIFF
--- a/src/finn/transformation/fpgadataflow/templates.py
+++ b/src/finn/transformation/fpgadataflow/templates.py
@@ -111,6 +111,9 @@ set_param board.repoPaths $paths_param
 if {$BOARD == "ZCU104"} {
     set_property board_part xilinx.com:zcu104:part0:1.1 [current_project]
     set ZYNQ_TYPE "zynq_us+"
+} elseif {$BOARD == "ZCU102"} {
+    set_property board_part xilinx.com:zcu102:part0:3.3 [current_project]
+    set ZYNQ_TYPE "zynq_us+"
 } elseif {$BOARD == "Ultra96"} {
     set_property board_part em.avnet.com:ultra96v1:part0:1.2 [current_project]
     set ZYNQ_TYPE "zynq_us+"


maltanar:
Thanks Felix!

FADU-Sam:
Have you ever ran tfc_end2end_example.ipynb with ZCU102 board setting?
image
If I run the notebook, there is a stitch IP synthesis error which was not generated with ZCU104 setting.

fpjentzsch:
I believe I tested this. Which error message are you seeing exactly?

This might be a Vivado license issue, as the ZCU102 part is not included in the WebPACK license.

FADU-Sam:
Yes, you are right. It is a license issue.
Is there a way to replace the license to a purchased one?
Since I set VIVADO_PATH to my tool location, I thought Vivado uses my license.

fpjentzsch:
Personally, I use the `XILINXD_LICENSE_FILE` environment variable to define the address of my license server, so I added the following line to my run-docker.sh script to set the same variable within the FINN docker environment:

`DOCKER_EXEC+="-e XILINXD_LICENSE_FILE=$XILINXD_LICENSE_FILE "`

FADU-Sam:
Hi Felix,

Thanks for your reply.
I added the following lines on the run-docker.sh(please find the attached
file).
: ${XILINXD_LICENSE_FILE="/home/samuel/Downloads"}  #location of my license
file
DOCKER_EXEC+="-e XILINXD_LICENSE_FILE=$XILINXD_LICENSE_FILE "

However it didn't solve the license issue.
Stitch IP synthesis failed with the same reason.
Since I am using a USB dongle key(MAC address) for my license, I doubt that
Docker can't see the USB dongle key properly.
Sorry not familiar with Docker.

Regards,
Samuel

2021년 4월 30일 (금) 오후 7:06, Felix Jentzsch ***@***.***>님이 작성:

maltanar:
Hi FADU-Sam, using USB dongle keys isn't something we support with FINN, but you may be able to work around this by either:

* generating all the necessary project files etc inside Docker, then running actual synthesis on them manually on the host (all generated files are normally on `/tmp/finn_dev_<username>` mounted both in container and on the host)
* trying to make the USB device directly visible inside the Docker container by adding more flags to the Docker launch commandline a la https://stackoverflow.com/questions/24225647/docker-a-way-to-give-access-to-a-host-usb-or-serial-device

FADU-Sam:
Hi Maltanar,

Thanks for your suggestions.
I ran actual synthesis on my host under /tmp/finn_dev_samuel and the bitstream generation was successful.
However I don't know how to skip the below process and run the next instruction.
image

I also tried to make USB device visible to the Docker by adding following flag to the Docker launch sh file.
DOCKER_EXEC+="--device=/dev/bus/usb/003/004 --volume=/dev/bus/usb/003/004:/dev/bus/usb/003/004 "
Unfortunately it didn't work out too(maybe it was not a right flag).
Since Docker allows to change MAC address of its network bridge, I added the below flag to the sh file.
DOCKER_EXEC+="--mac-address xx:xx:xx:xx:xx:xx " 
I expected the flag would make the Docker bridge having the same MAC address that my USB dongle has.
docker network inspect <NETWORK ID>  reported that MAC address correctly got replaced.
But it didn't solve the synthesis issue and got the same license complain.

Vivado expects to see its license under /home/<user>/.Xilinx according to Vivado license manager.
I think somehow I need to include my license inside Docker and make Vivado to search the license under a specific directory.

PR #284, opened 2021-02-21
Change in how Netron takes parameters

neilkimn:
**Issue:** 
Netron visualization does not work due to change in how Netron takes parameters for the address.

**Solution:**
Change parameters passed to Netron from two being distinct to a tuple of parameters.


maltanar:
Hi Neil, thank you for the PR! Since the fix is only applicable to newer Netron installs (it breaks in older ones if e.g. the Docker image is using a cached version) I added a `netron>=4.7.9` specifier to our Dockerfile.

PR #286, opened 2021-02-23
[absorb]: used find_consumers instead of find_consumer

mmrahorovic:
Used find_consumers instead of find_consumer. For non-linear graphs, we must take into account that the output of a node can be linked to the second input of the successive node (e.g. when the successive node is a 'join-node').


auphelia:
Thanks @mmrahorovic. This patched `AbsorbTransposeIntoMultiThreshold` to use `find_consumers()[0]` so it wouldn't bail on a fork Transpose. On current dev that transform handles the fork case deliberately instead: it guards and skips fork Transposes entirely, so the singular `find_consumer` lookup is safe. Closing as superseded, thanks for the original fix.

PR #287, opened 2021-02-23
Fix IODMA interface width selection

fpjentzsch:
In some cases, this bug caused an erroneous `intfWidth` setting of the input IODMA, resulting in a significant performance regression for memory-bound accelerators.


maltanar:
Thanks Felix!

PR #292, opened 2021-03-02
Added MoveTransposePastMultiThreshold transformation

mmrahorovic:
[reorder]: added a new transformation MoveTransposePastMultiThreshold.
[test_move_transpose_past_multithreshold]: added a test case for the new transformation.

### DIFF
--- a/src/finn/transformation/streamline/reorder.py
+++ b/src/finn/transformation/streamline/reorder.py
@@ -33,6 +33,7 @@ from onnx import TensorProto
 
 from finn.transformation.base import Transformation
 import finn.core.data_layout as DataLayout
+from finn.transformation.general import SortGraph
 from finn.transformation.infer_shapes import InferShapes
 from finn.transformation.infer_datatypes import InferDataTypes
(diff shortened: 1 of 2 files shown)


auphelia:
Thanks a lot for this contribution, @mmrahorovic, and apologies it sat open for so long.

Revisiting it for release triage: the mainline behavior here (move Transpose past MultiThreshold and set the MT data_layout) is now covered on dev by AbsorbTransposeIntoMultiThreshold in src/finn/transformation/streamline/absorb.py. We verified on the exact pattern from this PR's test (parallel Transpose → MultiThreshold branches into an Add): the existing transform resolves the [0,3,1,2] (NHWC→NCHW) direction identically.

The one piece this PR additionally handled is the [0,2,3,1] direction, which the current transform doesn't cover. We don't have a model/flow that produces that direction into a MultiThreshold today, so we're closing this as superseded rather than carrying it forward. If that case does come up, it's a small addition to the existing AbsorbTransposeIntoMultiThreshold (one extra perm branch), and this PR remains the reference for it.

Closing as superseded. Thanks again for the work.

PR #294, opened 2021-03-04
Fix AXI addressing for large runtime-writable weight arrays

fpjentzsch:
Set the proper BD address segment range based on the exposed address width.


maltanar:
Thanks Felix! I've changed the end2end tests a bit to have a testcase that covers this, so now LFC end2end tests are using runtime-writable weights instead of the TFC ones.

PR #295, opened 2021-03-04
Added MoveIdenticalOpBeforeFork transformation

mmrahorovic:
[reorder]: added a new transformation MoveIdenticalOpBeforeFork.
[test_move_identical_op_before_fork]: added a test case for the new transformation.

### DIFF
--- a/src/finn/transformation/streamline/reorder.py
+++ b/src/finn/transformation/streamline/reorder.py
@@ -1114,3 +1114,53 @@ class MoveIdenticalOpPastJoinOp(Transformation):
 class MoveTransposePastJoinAdd(MoveIdenticalOpPastJoinOp):
     def __init__(self):
         super().__init__(["Transpose"], ["Add"])
+
+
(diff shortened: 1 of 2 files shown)


auphelia:
Thanks for this one too, @mmrahorovic, and sorry for the long delay in getting back to it.

This transform is technically not superseded, there's no BeforeFork equivalent on dev. But the direction it moves in is the reason we're closing it. 
FINN streamlining is built to push operations downward (toward the outputs): the whole fork/join family on dev — MoveOpPastFork (op before a fork →  duplicated onto each branch) and MoveIdenticalOpPastJoinOp (identical branch ops → past the join), all move things downstream. MoveIdenticalOpBeforeFork is the one move that goes the other way, hoisting identical branch ops back upstream before the fork.

Rather than introduce a transform that runs counter to that convention (and could fight the downward moves during streamlining), we'd prefer to keep the flow single-directional. So we're closing this as out of line with the streamlining direction rather than as redundant.

Appreciate the contribution, thanks again.

PR #307, opened 2021-04-14
Make linking optional

surangamh:
Make the linking step optional with an argument passed to VitisBuild


maltanar:
I've clarified what the new flag does a little more in the docstring. Thanks Suranga!

PR #314, opened 2021-04-29
[VitisBuild] Apply SLR and memory bank constraints during Vitis Linking

quetric:
Currently FINN ops take attributes for SLR and memory bank but these are not applied when linking. This PR fixes this and adds some reasonable defaults for memory bank assignment when no explicit value has been requested. Because of this last change, this PR makes #246 redundant.

@maltanar I've added finn-experimental because I test this in conjunction with the partitioner. The changes here don't require finn-experimental.


maltanar:
Thanks Lucian. I'm actually a little reluctant to clone finn-experimental as part of the default FINN Docker as I normally think of that as something a power-user would install themselves, but so long as we don't have code in this repo that actually depends on finn-experimental this should be fine (or rather, let's try and see what happens :)).

PR #323, opened 2021-05-08
Resolved minor bug in minimize accumulator width

mmrahorovic:
Many thanks to @auphelia for helping resolve this.

Minimizing the accumulator width for the VVAU seems to have a bug in it. The logic for setting the accumulator DataType is as follows:
we have two numbers, tdt_min and tdt_max, that define our range [tdt_min, tdt_max]. If tdt_min >= 0, all numbers are unsigned and we find the appropriate DataType with the get_smallest_possible function according to tdt_max. If tdt_min < 0, the numbers must be represented in signed format. If abs(tdt_min)>tdt_max, we find the appropriate DataType with the get_smallest_possible function according to tdt_min. If abs(tdt_min)<=tdt_max, we find the appropriate DataType with the get_smallest_possible function according to -tdt_max.

Now, the bug happens due to the way how the get_smallest_possible method is implemented. Based on the largest value, the function checks for each datatype, starting from unsigned datatypes to signed, whether the number fits within the range. Thus in order to get the smallest possible datatype for signed numbers, you must supply a negative number. The problem here is that the range for positive valued numbers is smaller than the range of negative valued numbers for signed representation. Thus, in the edge case for signed numbers where tdt_max = 2^(n-1), the function will assign the wrong datatype (namely n-bits instead of n+1).

In other words, given a number tdt_max (which represents the largest weight), the get_smallest_possible method should check for each DataType whether tdt_max can be represented. So, for unsigned and signed formats respectively, it should check whether tdt_max is in [0, 2^n - 1] or [-2^(n-1), 2^(n-1)-1]. However, to consider a signed datatype, you must supply negative
numbers and therefore, the get_smallest_possible method only compares the number against the lower boundary of a format to determine whether the number can be represented. In other words, what the function is doing is checking whether -tdt_max >= -2^(n-1) to know whether n-bits are enough. Therefore, if tdt_max=2^(n-1), the function will return an n-bit signed datatype, while in fact we need (n+1)-bits to represent 2^(n-1) in signed format. 

The solution is to account for this and supply -(tdt_max+1) in the get_smallest_possible method. It follows when you write out the comparison actually performed in the get_smallest_possible method:
- for signed numbers when the largest positive value (tdt_max) is greater than the largest negative value (tdt_min), we want to confirm whether tdt_max fits within the range of n-bits by checking: tdt_max <= 2^(n-1)-1.
- what the get_smallest_possible method effectively does (for signed numbers) is comparing a number 'x' against -2^(n-1), i.e.
  x >= -2^(n-1).
- rewriting the equation above (the expression that we want) gives 
tdt_max        <= 2^(n-1)-1      ==
-tdt_max       >= -(2^(n-1)-1)  == 
-(tdt_max+1) >= -2^(n-1)
- thus, to get the right comparison, we must supply -(tdt_max+1) to the get_smallest_possible method.

Another solution would be to change the get_smallest_possible method in how it handles signed/unsigned formats (https://github.com/Xilinx/finn-base/blob/94beb27de0decb58d31555823860a24da5f09c5a/src/finn/core/datatype.py#L201).
For example to add an extra variable that specifies whether it should check for signed or unsigned numbers.


maltanar:
Hi @mmrahorovic , is there anything missing from this PR to be marked as ready? The reasoning in the explanation above is sound and I think the extra -1 is a good solution to this, but the PR is still marked as draft so wanted to double-check if there was something else you wanted to add here.

mmrahorovic:
Hi @maltanar, the PR is indeed ready, thank you for reminding me.

Issue #298, opened 2021-03-16
Network architecture - spatial dim in conv-> fc transition


dprzewlocka:
Hi,

following the issue mentioned in the gitter channel, I am facing the problem with custom network architecture (I'll repeat the description and add some new information).
My network is build of two conv layers and two fc layers (with bn, pool and activation between). During the streamlining, precisly the `AbsorbTransposeIntoMultiThreshold()` I got the assertation error `Untested spatial dim in conv->fc transition, proceed with caution!` (file finn/src/transformation/streamline/absorb.py, around line 355). In Brevitas/FINN examples the network CNV1w1a has subsequent layers build in such way, that the size of feature maps from the last convolutional layer is equal to 1. So I assumed this is the "tested spatial dim in conv->fc transition" - my network does not have this feature. So I commented out the assertation (as I decided to check it anyway). 
During hardware generation, I got the following error (file src/finn/transformation/fpgadataflow/insert_fifo.py, around line 98):
`The folded output shape of the first node is not the same as the folded output shape of the second node. A streaming fifo can't be implemented in between these nodes.`

After quick debugging it appeares that this situation is perhaps connected with layers last convolutional and first fully connected.
After the advice from @maltanar I decided to comment out this assertation and check what will happen. The hardware generation is now finishing without errors. I succesfuly run the network in the hardware, however output isn't correct.

I decided to check the Python simulation of the streamlined onnx model and the output matches the one obtained from hardware, but not the golden (Brevitas) one. What is more, the difference is already visible after first maxpool layer. I am following strictly the notebook cnv_end2end - that is, I also add the pre- and postprocessing (the Brevitas model accepts input in [0,1] range, the streamlined one includes division by 255).

I am not sure whether this is connected to this architecture issue, but until now I haven't found any solution. If it would be of any help, I can share the network's architecture. I'm also attaching the onnx model - the one obtained from the Brevitas model and the streamlined one (had to manualy add .zip to the name of the files, should work fine after renaming it back).

Thank you!

[XNOR-V3-T-Brevitas-Raw.onnx.zip](https://github.com/Xilinx/finn/files/6148705/XNOR-V3-T-Brevitas-Raw.onnx.zip)
[XNOR-V3-T-streamlined.onnx.zip](https://github.com/Xilinx/finn/files/6148769/XNOR-V3-T-streamlined.onnx.zip)

maltanar:
[...]

PR #326, opened 2021-05-11
Conv->FC transition w/ spatial dim != (1,1)

fpjentzsch:
Fixes issue #298.

Streamlining a network with a conv layer followed by a dense layer will now yield a (Transpose -> Flatten) pair in between. Since conv nodes operate on NHWC format, but the dense layer was trained under NCHW, this transpose will be absorbed into the dense weight matrix during convert_to_hls (RemoveCNVtoFCFlatten), which also removes the Flatten node.

Supports implicit flattening via a reshape node, e.g. reshape(1, -1).
Supports 1d convolutions as per #319, but the respective test cases have been disabled for now to ease merging.

**Depends on a minor fix in finn-base: https://github.com/Xilinx/finn-base/pull/30**
The commit pointer is still to be changed.


maltanar:
Thanks Felix!

PR #335, opened 2021-05-25
Support 1D MaxPool

fpjentzsch:
Adds support for the (non-binary) `StreamingMaxPool_Precision_Batch_1d` HLS op added in https://github.com/Xilinx/finn-hlslib/pull/49.

As with other recently added 1D layers, the data layout remains 4-dimensional (i.e. H or W set to 1).


maltanar:
Thanks for the PR @fpjentzsch -- but looking at the list of changes here after merging with latest `dev` it looks like this commit is undoing some changes from #332 (e.g. the extra `MakeMaxPoolNHWC` in `step_streamline` is removed). Is this intentional?

PR #347, opened 2021-06-14
Fixed input clock/reset port name inference during stitching

quetric:
When stitching, the BD global clock/reset input ports are produced by making external the clock/reset of the first IP in the dataflow. The stitching logic assumed that the name of those clock/reset signals on the first ip is ap_clk/ap_rst_n and if this is not the case, the stitching fails.

The fix utilizes the known port names to infer the correct BD global clock/reset names after making the ports external.

### DIFF
--- a/src/finn/transformation/fpgadataflow/create_stitched_ip.py
+++ b/src/finn/transformation/fpgadataflow/create_stitched_ip.py
@@ -120,13 +120,13 @@ class CreateStitchedIP(Transformation):
                 "make_bd_pins_external [get_bd_pins %s/%s]"
                 % (inst_name, clock_intf_name)
             )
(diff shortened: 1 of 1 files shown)


auphelia:
Thanks for this fix @quetric, and apologies it sat for so long.

The behaviour you added is now the default in the stitching flow. `CreateStitchedIP.connect_clk_rst` infers each node's clock/reset port names dynamically via `get_verilog_top_module_intf_names()["clk"][0]` / `["rst"][0]` and makes *those* pins external, rather than assuming the first IP's ports are named `ap_clk`/`ap_rst_n`. The `ap_clk`/`ap_rst_n` names that remain are just the renamed global BD ports.

Since the original issue (hardcoded first-IP clock/reset names) is fully covered, I'm closing this as already fixed. Thanks again for spotting and addressing it back then.

Issue #174, opened 2020-06-26
Caching support for long-running transformations


maltanar:
To speed up the compilation process for large models or large layers, it would make sense to have a caching mechanism for long-running transformations. The cached outputs would be persistent and get reused when appropriate when a transform is called. The cache generation/reuse should be optional. 

The idea would be to generate a hash for all relevant input data for a transform for a particular node (including node attributes, parameter tensors, quantization annotations...) and use that hash as the cache key for the output products in a folder in a persistent location. Later on, if the same transformation is executed on a node with the same hash, the cached outputs can be reused by copying from the persistent cache folder into a new folder.

For `NodeLocalTransform` the hashing is relatively straightforward as only the node itself is passed to the transformation. For others, it's hard to generalize, and best considered on a case-by-case basis that gives the most execution time benefits for the use cases we have.

@quetric @Tobi-Alonso do you have any suggestions for which transforms to start with to get the most benefit, or any other comments?

Tobi-Alonso:
For my test case using NUM_DEFAULT_WORKERS = 30 and StreamingFCLayer_Batch with "mem_mode" = "decoupled", these are the ones that take more time and I think are well suited for cache support:
- `HLSSynthIP` (1:40 hours) 
- `CompileCppSim` (~25 min from)  
- `PrepareIP`(23 min, this is still sequential and could easily be parallelized as done for `PrepareCppSim`), 
- `PrepareRTLSim`(8 min) 
- `PrepareCppSim`(~5 min)

I would start with `PrepareIP` and `HLSSynthIP`(needs cache support in`PrepareIP`).

PR #349, opened 2021-06-16
Implement IP caching

quetric:
This PR addresses #174 

I added the following features:

- two node attributes, `ip_cache_key` to store a key in an IP cache and 'disable_ip_synth' to instruct HLSSynthIP to skip a node
- an additional hlscustomop function , `get_structural_parameters()` which is meant to return a dict of the parameters which affect the generated hls; this dict can be hashed and turned into a cache key
- an additional hlscustomop function, `set_synth_enablement()` which the custom op can redefine if it wants to tell the HLSSynthIP to skip it for synthesis (for example for nodes which have HLS and RTL/Vivado implementations but are currently not configured to use the HLS)
- a new transform CacheIP which identifies unique IPs, synthesizes them, stores the relevant parameters into the cache, and adds a model attribute pointing to the cache file (a JSON)
- a test for the above

The test is by no means exhaustive but it does pass - graph is stitched in IPI with no issues. To be able to use the cache, each custom op should define the get_structural_parameters as appropriate for itself. By default nodes return no parameters and are therefore always synthesized from scratch. I implemented get_structural_parameters for FC layers only for this test.

### DIFF
--- a/src/finn/custom_op/fpgadataflow/convolutioninputgenerator.py
+++ b/src/finn/custom_op/fpgadataflow/convolutioninputgenerator.py
@@ -88,6 +88,23 @@ class ConvolutionInputGenerator(HLSCustomOp):
         my_attrs.update(super().get_nodeattr_types())
         return my_attrs
 
+    def get_structural_parameters(self):
+        # node is cache-able, collect parameters influencing HLS
+        # any array of ints must be converted to tuple so the dict is hashable
+        ret = dict()
+        ret["ConvKernelDim"] = tuple(self.get_nodeattr("ConvKernelDim"))
+        ret["IFMChannels"] = self.get_nodeattr("IFMChannels")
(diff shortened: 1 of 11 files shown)


auphelia:
Thanks for this @quetric, the idea (synthesize one IP per unique layer configuration and reuse it across identical layers, instead of re-synthesizing) is still valuable.

However, this draft has been open since 2021 and never landed, and the current codebase has moved on: there's no equivalent cross-layer caching today (only per-node skip-if-exists in prepare_ip.py and hlssynth_ip.py, which avoids re-work for the same node instance on a rerun). Landing the structural-parameter hashing + `CacheIP` transform + `ip_cache_key`/`disable_ip_synth` design would require a substantial rewrite against today's `PrepareIP`/`HLSSynthIP` rather than a rebase.

Closing as stale. Happy to revisit the feature as a new PR against the current flow if it becomes a priority.

PR #351, opened 2021-06-18
Disable casting individual weight values

mmrahorovic:
Write individual weight values as integer values in the weight array in params.h for a VVAU. This lowers the HLS synthesis time.

### DIFF
--- a/src/finn/custom_op/fpgadataflow/vector_vector_activate_batch.py
+++ b/src/finn/custom_op/fpgadataflow/vector_vector_activate_batch.py
@@ -312,7 +312,8 @@ class Vector_Vector_Activate_Batch(HLSCustomOp):
         code_gen_dir = path
 
         """Saves weights into params.h"""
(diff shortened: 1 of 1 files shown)


auphelia:
Thanks @mmrahorovic, and sorry this sat so long. Closing as no longer applicable against current dev:

The PR drops the innermost SIMD dim (shape[:-1]) and emits unpacked weights, which is only valid when SIMD == 1 (true for VVAU in 2021). Current VVAU_hls supports SIMD > 1 kernel folding, where SIMD lanes must be packed into one ap_uint<SIMD*W> word, so the unpacked form is now incompatible. VVAU also mirrors MVAU's packed weight-gen path.

The original motivation (faster HLS elaboration) is also largely addressed elsewhere now: internal_decoupled mem_mode already cuts elaboration time significantly by streaming weights instead of baking them into the header, and the RTL variants have no HLS synthesis at all. So I'd just drop this one.

PR #353, opened 2021-06-23
Added SIMD requirement to StreamingFCLayer_Batch

HenniOVP:
Hi,
the added SIMD check can be found in line 1015 to 1024.
The pre-commit hooks also changed some of the other lines.


maltanar:
Thanks @HenniOVP !

PR #354, opened 2021-06-25
Feature/pre commit adherence

HenniOVP:
Hi,
I updated the pre-commit configuration to be on par with finn-base and then executed it on the whole code-base of FINN.
Unfortunately, I had to remove `debug-statements` hook, because it ignores `# NOQA` commands. Which means that no commit, which changes the `src/finn/builder/build_dataflow.py` file, would pass the `debug-statements` pre-commit hook.

Additionally, I had to resolve some issues, which are potentially critical. These are in commit hash 51b95ca1d60e230a513d6643a9e7bc3bfe3400d4 and it would be great if someone could take a closer look at those changes.

As a general aside, I also noticed that all `*.v` file have CRLF line-endings, but I'm not sure if that's actually an issue.


src/finn/transformation/fpgadataflow/floorplan.py
```diff
@@ -70,7 +70,7 @@ def apply(self, model):
 
         try:
             default_slr = self.user_floorplan["Defaults"]["slr"][0]
-        except:
+        except Exception:
```
HenniOVP, line 73: @maltanar I forgot to mention this change, which was required by the pre-commit hooks. This change will still catch almost all exceptions, excluding: `SystemExit`, `KeyboardInterrupt` and `GeneratorExit`. And I don't think we really want to catch those excluded exceptions here.

maltanar: Good catch (pun intended) -- here we would primarily be interested in the cases where there's no specified floorplan entry. I think what the original author (@quetric ) is trying to capture here is that the exception may be of various types (e.g. `user_floorplan` may be None, or it may not have the defaults set up). Let's go with this solution.

maltanar:
Thanks @HenniOVP -- I've run the FINN test suite on this and looks like none of the reformatting broke anything. Since we want to keep to exit-to-pdb behavior for failed builds, I'm OK with the removal of `debug-statements:` -- we'll just have to manually watch out for additional pdb entries.

HenniOVP:
Nice, great that everything seems to still work as expected.
Concerning the `debug-stamets` hook: My later commit 8175b9b actually added the hook back in, but it now explicitly excludes the `src/finn/builder/build_dataflow.py` file. With this we won't have to manually watch all new pdb entries anymore, but just the one in `src/finn/builder/build_dataflow.py` and where the file moves, in case it ever does. Sorry for not making that clear.

PR #364, opened 2021-07-26
Non-square thresholding

quetric:
These changes enable easier manipulation of thresholding blocks for non-square tensors and working with MMV.

Changes:

- moved streaming thresholding upstream to hlslib
- modified templates for thresholding blocks to take image height and width instead of assuming square images
- modified FINN to use the updated hlslib and to generate proper calls (this PR)


HenniOVP (approved):
Looks good! Everything seems to be implemented in a pretty straightforward way.

maltanar:
`test_runtime_thresholds_single_layer` seems to be currently failing, though all other rtlsim/cppsim tests for this layer pass. I'll need to take a look as the test also seems to be failing in `dev`.

maltanar:
Update: `test_runtime_thresholds_single_layer` passes on 2019.1 but fails on 2020.2 so there is some Vivado HLS version dependency here. Since this is independent from this PR I will merge this as-is and create a new issue to track the runtime thresholds problem.

maltanar (approved):
LGTM

PR #324, opened 2021-05-10
Updated shownInNetron function

pinxau1000:

PR #366, opened 2021-08-06
Added support for externally set netron port, this resolves #365 and can close #324

### LINKED ISSUE #365: showInNetron function dosn't use custom netron port set as enviroment variable.
This requires a few changes to the function itself here: https://github.com/Xilinx/finn/blob/bdfbd4b79088accf92e60fc1fe790e697500dfe7/src/finn/util/visualization.py#L39

And a way to get the port number itself, likely similar to this: https://github.com/Xilinx/finn-base/blob/c928353369f9f4c76e6bbd5c2d0f68c20763c44b/src/finn/util/basic.py#L106

HenniOVP:
This PR adds support for an externally set netron `port`, either via the environment variable or as a function parameter, same for the `localhost_url`.
Additionally, I adapted the doc string from the PR #324 to adhere to the Sphinx formatting: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format

### DIFF
--- a/src/finn/util/visualization.py
+++ b/src/finn/util/visualization.py
@@ -36,7 +36,29 @@ def showSrc(what):
     print("".join(inspect.getsourcelines(what)[0]))
 
 
-def showInNetron(model_filename):
-    netron.start(model_filename, address=("0.0.0.0", 8081))
-    localhost_url = os.getenv("LOCALHOST_URL", default="localhost")
-    return IFrame(src="http://%s:8081/" % localhost_url, width="100%", height=400)
+def showInNetron(model_filename: str, localhost_url: str = None, port: int = None):
+    """Shows a ONNX model file in the Jupyter Notebook using Netron.
+
+    :param model_filename: The path to the ONNX model file.
+    :type model_filename: str
+
+    :param localhost_url: The IP address used by the Jupyter IFrame to show the model.
+     Defaults to localhost.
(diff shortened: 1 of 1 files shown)


src/finn/util/visualization.py
```diff
@@ -36,7 +36,29 @@ def showSrc(what):
...
+    :type port: int, optional
+
+    :return: The IFrame displaying the ONNX model.
+    :rtype: IPython.lib.display.IFrame
+    """
+    if port is None:
+        try:
+            port = int(os.environ["NETRON_PORT"])
+        except KeyError:
+            port = 8081
```
rpitonak, line 60: You can possibly re-write this to:
```
port = port or int(os.getenv("NETRON_PORT",  "8081"))
```
And maybe catch an int conversion exception instead.

rpitonak (commented):
Hello! Thanks for this feature, I added one suggestion for refactoring

src/finn/util/visualization.py
HenniOVP: Thanks for the suggestion, that's a really good idea! I refactored the code accordingly and did the same adjustment for the `localhost_url` argument.

maltanar:
Looks good, thanks @HenniOVP and @rpitonak!

PR #369, opened 2021-08-16
Fixed off by one error in SIMD check for StreamingFCLayer_Batch.

HenniOVP:
Hi,
unfortunately I had a typo in the last SIMD check, which resulted in a off by one error.
In actuality buffers with a size of 1024 are still possible as one can see in the log of the original issue here: https://forums.xilinx.com/t5/High-Level-Synthesis-HLS/ARRAY-PARTITION-COMPLETE-has-exceeded-the-threshold-1024/td-p/862145


maltanar:
LGTM and well spotted, thanks @HenniOVP !

PR #373, opened 2021-09-17
Blog entry for FINN MVU RTL implementation

asadalam:
The blog entry briefly describes the MVU RTL implementation and also incorporates the README file from the main repository.


maltanar:
Many thanks @asadalam ! I just added a small note that we plan to do the integration of this into the FINN compiler in the near future, otherwise good to go.

PR #384, opened 2021-10-08
Support for ingesting models in the QONNX format

heborras:
### Summary

This PR is designed to add support for the ingestion of the QONNX format in FINN.
In particular a new transformation (`ConvertQONNXtoFINN`), a corresponding step in the dataflow builder (`step_qonnx_to_finn`) and an optional flag (`expect_QONNX_as_input`) have been added.

### Details

The PR exchanges the finn-base installation for the qonnx front end, which implicitly installs the `qonnx_quant_op` branch of finn-base.

The conversion from QONNX to FINN-ONNX is implemented in the following steps:
- Convert `Gemm` node to `MatMul` and `Add` node (The `Gemm` node is exported by the mobilenet and constant folding the quantized weights requires this node to be split up.)
- Constant fold weights, which are expressed as Quant nodes
- Convert Quant nodes in the activation path to MultiThreshold nodes

### Tests

Additionally a new testcase (`test_QONNX_to_FINN`) has been added to verify that all Quant nodes get converted by this transformation and that the output during execution is still equivalent to the FINN-ONNX exported by Brevitas.

The test currently examines the following networks:
- TFC, SFC (all W2A2)
- CNV W2A2
- Mobilenet (except for Trunc node, see ToDo list)

With this the test in particular covers the following activation types:
- ReLu
- Idendity

And the following weight types:
- Integers above 1 bit
- Channel-scaled integers of four bit

### Open ToDos 
These ToDos need to be addressed before review and merging can take place.
- [x] Add an end2end test to verify that the rest of the FINN toolchain still works as expected with the generated FINN-ONNX model.
- [x] Add support for converting the `AvgPool` + `Trunc` node construct to a `QuantAvgPool2d` node. This in particular requires a patch to the Brevitas export of the `Trunc` node, as the node currently dosn't contain enough information to be executed.
- [x] Add support for binary Quant nodes, which are planned to get introduced in the Brevitas export along with the patched `Trunc` node.

### DIFF
--- a/tests/fpgadataflow/test_fpgadataflow_ipstitch.py
+++ b/tests/fpgadataflow/test_fpgadataflow_ipstitch.py
@@ -44,7 +44,6 @@ from finn.transformation.fpgadataflow.floorplan import Floorplan
 from finn.transformation.fpgadataflow.hlssynth_ip import HLSSynthIP
 from finn.transformation.fpgadataflow.insert_iodma import InsertIODMA
 from finn.transformation.fpgadataflow.insert_tlastmarker import InsertTLastMarker
-from finn.transformation.fpgadataflow.make_deployment import DeployToPYNQ
 from finn.transformation.fpgadataflow.make_zynq_proj import ZynqBuild
 from finn.transformation.fpgadataflow.prepare_ip import PrepareIP
 from finn.transformation.fpgadataflow.synth_ooc import SynthOutOfContext
@@ -363,11 +362,6 @@ def test_fpgadataflow_ipstitch_zynqbuild(board):
         assert sdp_node.__class__.__name__ == "StreamingDataflowPartition"
         assert os.path.isfile(sdp_node.get_nodeattr("model"))
         model = load_test_checkpoint_or_skip(sdp_node.get_nodeattr("model"))
-    # generate inputs for remote exec
-    iname = "inp"
-    idt = model.get_tensor_datatype(iname)
-    ishape = model.get_tensor_shape(iname)
-    x = gen_finn_dt_tensor(idt, ishape)
     # bitfile using ZynqBuild
     model = model.transform(ZynqBuild(board, 10))
     model.save(ip_stitch_model_dir + "/test_fpgadataflow_ipstitch_customzynq.onnx")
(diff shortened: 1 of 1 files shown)


HenniOVP:
The feature has now matured so far, that I've started adding it to more and more tests in the test suite. And this has already lead to a few bugs being discovered and fixed.

So the whole list of tests, which currently test the QONNX ingestion, are the following existing or new tests:

- New:
  - `test_QONNX_to_FINN`
- Existing:
  - `TestEnd2End`
    - This is the bnn_pynq end2end test.
  - `test_brevitas_cnv_export_exec`
  - `test_brevitas_fc_onnx_export_and_exec`
  - `test_brevitas_act_export_qhardtanh_nonscaled`
  - `test_brevitas_QConv2d`
  - `test_brevitas_qlinear`
  - `test_brevitas_act_export_relu`
  - `test_brevitas_act_export_qhardtanh_scaled`

The following tests are still under development and are so to say ToDos:
- [ ]  `test_end2end_cybsec_mlp_`*
  - Development is still work in progress for these tests.
- [ ] `test_brevitas_debug` 
  - Support for this just got added today in Brevias, so I simply haven't gotten around to look at it.
- [ ] `test_brevitas_avg_pool_export`
  - This test needs significant changes in design due to how the export products with quantized input tensors vary greatly between the FINNManager and the BrevitasONNXManager. So there is significantly more work that needs to be done here and I'm simply not yet finished with it.
- [ ]  `test_brevitas_act_export_relu_imagenet`
  - Similar-ish issues as with `test_brevitas_avg_pool_export`.

And finally these tests are currently "won't add" due to different reasons:
- test_brevitas_mobilenet
  - The test is currently marked as xfail and a similar case is already tested by the test_QONNX_to_FINN test.
- test_brevitas_compare_exported_mobilenet
  - For the same reason as for the test above.
- test_end2end_mobilenet_v1*
  - This is technically also because some of the tests are marked as xfail, but it might still be interesting to add these tests nonetheless.
- test_ext_weights*
  - Here the .onnx file is loaded externally and not exported, so it's not easily adaptable.

HenniOVP:
@maltanar the most recent commit 6a22a11fee35a61bea43ab0ad2f58db53cfb68c3 is a bug fix for an issue, which Ben had with the Pynq-Z2. Just to make sure, that you're not surprised by why I changed that file.

maltanar:
All good to go, thanks @HenniOVP !

PR #412, opened 2021-11-05
First draft for QONNX blog post.

HenniOVP:
Hi @maltanar,
Here is the first draft for the blog post on QONNX. I tried to build the post such that I first explain the motivation of why QONNX is interesting for FINN before going into how QONNX actually looks like and how to use it.

It would be great if you could take a look at the post and add comments and maybe additions if anything comes to mind.


maltanar (approved):

maltanar:
Thanks a lot @HenniOVP ! I've done some edits to the verbosity of the text and the order of sections.

PR #416, opened 2021-11-08
Support stride > 1 for parallel 1D SWG

fpjentzsch:
**Depends on:** https://github.com/Xilinx/finn-hlslib/pull/65


auphelia:
As discussed with the author (@fpjentzsch), this functionality is now covered by the RTL SWG implementation (and potentially newer HLS SWG variants as well).

Since we do not plan to maintain multiple implementations of the same feature, we'll close this PR. Thanks for the contribution and for the discussion.

PR #464, opened 2021-12-06
Support for 1D SWU

mmrahorovic:


maltanar:
Thanks for the PR @mmrahorovic , couple of comments on this one:
* I'm guessing this PR should be merged after https://github.com/Xilinx/finn-hlslib/pull/68 ?
* Since the hlslib PR is targeting Vitis HLS, the target branch for this PR should be `feature/vitis_hls_2021.2`
* `ConvolutionInputGenerator_1D_dilated_dws` seems to be missing from the hlslib-side PR - I could only see `ConvolutionInputGenerator_NonSquare_Dilated` in `slidingwindow.h`

Would be great if we can also update the resource estimates, though this can be left for further study and a future PR.

PR #491, opened 2022-01-13
GPU training support for the cybersecurity MLP notebook

volcacius:


maltanar:
Thanks @volcacius ! The CI check failing above is a known issue with the random seed for MobileNet and is unrelated. I tested this for both CPU and GPU modes and seems to be working well, so merging now.

PR #607, opened 2022-06-13
RTL ConvolutionInputGenerator

fpjentzsch:
## Overview
Adds an RTL Convolution Input Generator / Sliding Window Generator (SWG) that aims to replace all 16 "ConvolutionInputGenerator_*" [function variants in finn-hlslib](https://github.com/Xilinx/finn-hlslib/blob/vitis-hls/slidingwindow.h), extend functionality, and improve resource & latency efficiency.

Depending on the folding configuration, HDL code for one of two implementation styles is generated via (System-)Verilog templates.

- **Out width <= in width**: Use an addressable cyclic buffer.
- **Out width > in width**: Use a combination of shift registers (which can be read in parallel) and cyclic line buffers.

Despite the RTL implementation, the layer is provided as an **HLSCustomOp (!)**. More details to come...
### This is still a work-in-progress, required TODOs before merge:
- [ ] Additional testing
- [ ] Update cycle & resource estimation
- [ ] Ensure proper auto folding
- [ ] Code cleanup and optimization

### There are also some optional features that could be left for a future PR:
- [ ] Support MMVin > 1 for MMVout = 1 and MMVout = K
- [ ] Support M > 1 for the general 2D case (alongside "MMV" support for MVAU, etc.)


auphelia:
Hi @fpjentzsch, could you change the target branch for the merge to target `dev`, please?

fpjentzsch:
Hi @preusser, @maltanar, @auphelia, 
as discussed, I moved the development of the dynamic FM sizing and parallel implementation mode to different branches for now so we can merge this part first.

finn-rtllib/swg/swg_template_default.sv
```diff
@@ -0,0 +1,335 @@
...
+    assign  tail_incr = tail_incr_reg;
+
+    // combinational logic for tail_incr generation
+    uwire tail_incr_inner_condition;
+    generate
+        if (IS_DEPTHWISE)
+            assign tail_incr_inner_condition = (Counter_loop_kh >= 0);
+        else
+            assign tail_incr_inner_condition = 0;
+    endgenerate
```
preusser, line 50: `generate` blocks are legacy Verilog. Just remove them.
You can even opt for a concise single line also incorporating the declaration:
```
uwire  tail_incr_inner_condition = IS_DEPTHWISE? (Counter_loop_kh >= 0) : 0;
```

finn-rtllib/swg/swg_template_default.sv
```diff
@@ -0,0 +1,335 @@
...
+
+    always @ (tail_incr_inner_condition, Counter_loop_w, Counter_loop_h) begin
```
preusser, line 52: This arguably does _not_ describe a register. So, please, revise the signal name `tail_incr_reg`.
Legacy `always @(...)` blocks should also be avoided unless there is a very good excuse for using them. In this case, an `always_comb` appears to be the preferable substitute.

finn-rtllib/swg/swg_template_default.sv
```diff
@@ -0,0 +1,335 @@
...
+    input   logic  read_enable,
+    input   logic [$clog2(DEPTH)-1:0]  read_addr, // absolute (!) read address of cyclic buffer
+    output  logic [WIDTH-1:0]  data_out
+);
+
+    $RAM_STYLE$ logic [WIDTH-1:0] Ram[DEPTH];
+    logic [WIDTH-1:0]  Out = 'x;
+    always_ff @(posedge clk) begin
+        if (!rst_n) begin
+            Out       <= 'x;
```
preusser, line 136: Yes, it is good style to mark dont-cares upon reset explicitly.
Unfortunately, this pattern regularly prevents Vivado from merging a memory read output register into a BRAM or URAM primitive. In this particular circumstance, it is, therefore, helpful not to mention the reset at all in the describing `always_ff` block.

finn-rtllib/swg/swg_template_default.sv
```diff
@@ -0,0 +1,335 @@
...
+    );
+
+    // Counters/address registers
+    // Add a sign bit even to (most) unsigned counters and Window_buffer_read_addr_reg,
+    // so we can use automatic sign extension and simplify calculations w/ signed increment.
+    // Alternatively, we could manually sign-extend and shave off a bit here or there.
+    logic signed [$clog2(LAST_READ_ELEM+1)+1-1:0]  Newest_buffered_elem = -1;
+    logic        [$clog2(LAST_READ_ELEM+1)+1-1:0]  Current_elem = 0;
+    logic        [$clog2(LAST_READ_ELEM+1)+1-1:0]  First_elem_next_window = 0;
+    logic        [$clog2(ELEM_PER_WINDOW)   -1:0]  K = 0;
```
preusser, line 216: It'd be helpful if you could come up with a somewhat longer name. `K` simply looks too much like some constant parameter.

finn-rtllib/swg/swg_template_wrapper.v
```diff
@@ -0,0 +1,54 @@
+`timescale 1 ns / 1 ps
+
+module $TOP_MODULE_NAME$ (
```
preusser, line 3: Is there a reason to use legacy Vivado module interface syntax?

preusser (changes requested):

src/finn/transformation/fpgadataflow/set_folding.py
```diff
@@ -174,6 +175,7 @@ def apply(self, model):
                 if op_type in [
```
maltanar, line 175: minor: in L159 this is used instead of listing the variants explicitly:
`if swu_node.op_type.startswith("ConvolutionInputGenerator"):`

suggest to either use that style here, or change L159 to also use the explicit listing to keep both of them consistent.

src/finn/transformation/fpgadataflow/convert_to_hls_layers.py
```diff
@@ -128,105 +132,130 @@ def apply(self, model):
...
-                    )
-                    assert is_equal_stride, (
-                        """%s : DownSampler currently only supports equal stride value
-                        along different axes."""
-                        % n.name
-                    )
-                    ConvInpGen_idim = ConvInpGen_idim_h
-                    stride = stride_h
-                    # create DownSampler node
+                if self.use_rtl_variant:
```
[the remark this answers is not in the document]
maltanar: e.g. `test_convert_to_hls_conv_layer` seems to suggest 1x1 kernels are not supported?
```
    if use_rtl_swg and kernel_size == 1:
        pytest.skip("1x1 kernel not supported by current RTL SWG")

```

tests/fpgadataflow/test_convert_to_hls_conv_layer.py
```diff
@@ -156,10 +163,15 @@ def test_convert_to_hls_conv_layer(conv_config, depthwise, exec_mode):
...
     inp_dict = {model.graph.input[0].name: x}
     assert oxe.compare_execution(model, new_model, inp_dict)
+
+    if use_rtl_swg:
+        downsampler_op_type = "ConvolutionInputGenerator_rtl"
+    else:
+        downsampler_op_type = "DownSampler"
     if kernel_size == 1 and stride > 1 and pad == 0:
-        assert new_model.graph.node[1].op_type == "DownSampler"
+        assert new_model.graph.node[1].op_type == downsampler_op_type
```
maltanar, line 172: maybe I missed something here but there is a skip condition above:

```
    if use_rtl_swg and kernel_size == 1:
        pytest.skip("1x1 kernel not supported by current RTL SWG")
```

if the 1x1 kernel is not support by the RTL SWG and its testcase will be skipped, does this piece of code ever get triggered?

tests/fpgadataflow/test_fpgadataflow_convinputgenerator_rtl.py
```diff
@@ -0,0 +1,259 @@
...
+def prepare_inputs(input_tensor):
+    return {"inp": input_tensor}
+
+
+# input datatype
+@pytest.mark.parametrize("idt", [DataType["UINT4"]])
+# kernel size
+@pytest.mark.parametrize("k", [[2, 2], [3, 3], [1, 3]])
+# input dimension
+@pytest.mark.parametrize("ifm_dim", [[24, 24], [13, 13], [1, 14]])
```
maltanar, line 145: suggest to add a rectangular (nonsquare 2D) testcase to the configs here

src/finn/custom_op/fpgadataflow/convolutioninputgenerator_rtl.py
```diff
@@ -0,0 +1,817 @@
+# Copyright (c) 2020, Xilinx
```
maltanar, line 1: copyright year

src/finn/custom_op/fpgadataflow/convolutioninputgenerator_rtl.py
```diff
@@ -0,0 +1,817 @@
...
+
+    def get_nodeattr_types(self):
+        my_attrs = {
+            "ConvKernelDim": ("ints", True, []),  # [H, W] = [Y, X]
+            "IFMChannels": ("i", True, 0),
+            "IFMDim": ("ints", True, []),  # [H, W] = [Y, X]
+            "OFMDim": ("ints", True, []),  # [H, W] = [Y, X]
+            "SIMD": ("i", True, 0),
+            "M": ("i", False, 1),
+            "parallel_window": ("i", False, 0, {0, 1}),
```
maltanar, line 75: can set allowed values to be only `{0}` to restrict no parallel window mode in this version

src/finn/custom_op/fpgadataflow/convolutioninputgenerator_rtl.py
```diff
@@ -0,0 +1,817 @@
...
+                0,
+                ((stride_w - 1) - (int(mmv_out * k_h * k_w / mmv_in))) * channel_factor,
+            )
+            + max(
+                0,
+                ((stride_h - 1) * w - (int(mmv_out * k_h * k_w / mmv_in)))
+                * channel_factor,
+            )
+        )
+        self.buffer_depth = buffer_actual_size  # for resource estimation
```
maltanar, line 477: it isn't safe to assume that ordinary Python member variables stick around as part of `CustomOp` instances, since they get may get serialized to ONNX and then de-serialized back for e.g. the next `build_dataflow` step. 

so this `buffer_depth` should be either made into a member function that gets computed dynamically based on other attributes or made into an attribute itself. thus it becomes a part of the ONNX node attribute system and gets serialized/deserialized safely.

maltanar (changes requested):

finn-rtllib/swg/swg_template_wrapper.v
fpjentzsch: Do you mean the declaration of port direction and Vivado interface params directly in the port list, like so?

```suggestion
module $TOP_MODULE_NAME$ (
(* X_INTERFACE_PARAMETER = "ASSOCIATED_BUSIF in0_V:out_V" *)
input  ap_clk,
(* X_INTERFACE_PARAMETER = "ASSOCIATED_BUSIF in0_V:out_V" *)
input  ap_rst_n,
input  [BUF_IN_WIDTH-1:0] in0_V_TDATA,
input  in0_V_TVALID,
output in0_V_TREADY,
output [BUF_OUT_WIDTH-1:0] out_V_TDATA,
output out_V_TVALID,
input  out_V_TREADY
);
```

If you're asking why this is not SystemVerilog: Vivado doesn't allow it as top-level for RTL modules.

tests/fpgadataflow/test_convert_to_hls_conv_layer.py
fpjentzsch: You're right, this is a remnant of the "parallel mode" that supported this case. I removed this piece of code for now.

src/finn/transformation/fpgadataflow/convert_to_hls_layers.py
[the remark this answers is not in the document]
fpjentzsch: I added a check for the 1x1 kernel case (which will be supported later using "parallel mode") so that the conversion falls back to HLS with a warning.

Other existing checks or skip conditions target generally-illegal convolution hyperparameters or unsupported folding factors, both of which are not really appropriate to check during HLS conversion.

fpjentzsch:
Hi @preusser, @maltanar,
thanks for your comments and tips! I incorporated them in above commit and marked those that should not require further discussion as resolved.

maltanar (approved):
LGTM, thanks Felix!

mgehre-amd:
This looks nice! Do you have numbers comparing resource utilization/latency between the C++ and RTL implementations?

finn-rtllib/swg/swg_template_wrapper.v
preusser: Looks better this way, thanks!

preusser (approved):

fpjentzsch:
Hi @mgehre-amd, thanks, I plan to (re-)generate some experiments for a comprehensive HLS/RTL comparison once I find some time. I'll get back to you and link the results here.

PR #615, opened 2022-06-22
Feature/checksum test

auphelia:
Fix bug in insert_hook transform and add FINN test for checksum layer

### DIFF
--- a/src/finn/transformation/fpgadataflow/insert_hook.py
+++ b/src/finn/transformation/fpgadataflow/insert_hook.py
@@ -122,6 +122,7 @@ class InsertHook(Transformation):
                             else:
                                 model.graph.output.pop()
                                 model.graph.output.append(chk_otensor)
+                                model.graph.value_info.remove(chk_otensor)
                                 model = model.transform(GiveUniqueNodeNames())
                                 model = model.transform(GiveReadableTensorNames())
                             graph_modified = True
--- a/src/finn/transformation/fpgadataflow/insert_hook.py
+++ b/src/finn/transformation/fpgadataflow/insert_hook.py
@@ -37,7 +37,7 @@ from finn.util.fpgadataflow import is_fpgadataflow_node
 
 
 def _is_hook_node(node):
-    if node.op_type in ["checksum"]:
+    if node.op_type in ["CheckSum"]:
         return True
     else:
         return False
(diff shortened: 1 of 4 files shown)


src/finn/transformation/fpgadataflow/insert_hook.py
```diff
@@ -37,7 +37,7 @@
 
 
 def _is_hook_node(node):
-    if node.op_type in ["checksum"]:
+    if node.op_type in ["CheckSum"]:
```
mmrahorovic, line 40: If this is the only 'hook' type node, then we could do the check like: `if node.op_type == "CheckSum"`. 
Perhaps in the future, if we plan to extend the types of 'hook_nodes', we could have a single class HookNode which has an attribute (e.g. 'checksum') which specifies what it does.

src/finn/transformation/fpgadataflow/insert_hook.py
```diff
@@ -82,7 +82,7 @@ def apply(self, model):
                     if n0_hook in list_supported_hooks:
```
mmrahorovic, line 82: Currently there is only 1 hook in the `list_supported_hooks`, so this if-branch is always taken (and the if-statement on the next line is exactly the same).
But again, for the future if we would plan to extend this, it makes sense.

mmrahorovic, line 82: Related to line 74.

Just listing this here, I think you have a better idea what exactly the CheckSum layer should perform.
The only node that can have multiple outputs is the DuplicateStreams_Batch. What will happen is that this transform will insert 2 CheckSum nodes for 2 identical streams (going out of the DuplicateStreams_Batch), which perhaps is overkill. If the layer preceding the DuplicateStreams_Batch has a CheckSum node, then perhaps we don't need to compute it afterwards again.

tests/fpgadataflow/test_fpgadataflow_checksum.py
```diff
@@ -0,0 +1,201 @@
...
+test_fpga_part = "xczu3eg-sbva484-1-e"
+target_clk_ns = 5
+
+
+def create_two_fc_model():
+    # create a model with two MatrixVectorActivation instances
+    wdt = DataType["INT2"]
+    idt = DataType["INT32"]
+    odt = DataType["INT32"]
+    m = 4
```
mmrahorovic, line 61: Maybe you could place these variables (wdt, idt, odt, m, actval, no_act, pe, simd) as a parameter to the test case. Then we could extend the test suite easier in the future and make it more visible which parameters can be altered.

However, this is not really important for now since this test already is sufficient.

tests/fpgadataflow/test_fpgadataflow_checksum.py
```diff
@@ -0,0 +1,201 @@
...
+    actval = 0
+    no_act = 1
+    binary_xnor_mode = 0
```
mmrahorovic, line 64: Not entirely sure, but you can leave `binary_xnor_mode` out since it will be `0` by default (and should be only set if weights/activations are bipolar if I'm not mistaken).

mmrahorovic (approved):

PR #642, opened 2022-07-15
[actions] Update quicktest workflow to new docker setup

auphelia:
Add env var for FINN_ROOT to quicktest workflow


auphelia:
Update quicktest workflow to match new Docker setup

PR #675, opened 2022-08-26
VVAU: add weightstream width helper

azizb-xlnx:
add the missing helper function to VVAU


maltanar:
Looks good, thanks Aziz!

PR #685, opened 2022-09-20
Support Vivado AXIS infrastructure IP in rtlsim

maltanar:
_Note: This PR depends on / includes PR #653  to get the `ap_none` interface support for FIFO level monitoring in hardware and PR #649 since it includes code to handle decoupled VVAU FIFO sizing correctly._

Originally, it was not possible to run pyverilator rtlsim on models that contain DWCs or FIFOs using `impl_style="vivado"` which makes use of the Vivado AXIS infrastructure components. For those cases, any incompatible DWCs/FIFOs had to be regenerated using the rtl/hls implementation styles, sometimes causing other compatibility issues (e.g. the HLS DWC requires widths to be divisible with each other, while the Vivado DWC relaxes this constraint).

This PR introduces a set of changes that makes it possible to use such components in rtlsim by:
- Upgrading to Verilator v4.224 to get some new features/bugfixes
- Including the necessary Verilog header files from Vivado as part of `pyverilate_stitched_ip` (including a modified copy of axis_infrastructure.vh with a non-recursive implementation of the gcd function)
- A new config option in `DataflowBuildConfig` called `rtlsim_use_vivado_comps=True` to switch between the new and old behaviors.

Additional fixes/improvements:
- Since this version of verilator breaks how pyverilator accesses internal signals: introduce tracking of max observed FIFO level since reset in `Q_srl.v` directly in hardware (port name `maxcount`, enabled with the `depth_monitor` attribute on `StreamingFIFO` nodes), and use this info in `InsertAndSetFIFODepths`. The transformation should also run faster as a result, since we don't have to monitor FIFO depths in the Python sim loop anymore.
- Use tensor size instead of constant depth in `InsertAndSetFIFODepths` by default. There is no guarantee the FIFO sizing will be done correctly for large image sizes otherwise.
- When inserting DWCs, automatically use Vivado mode if i/o widths aren't divisible


preusser:
Which parameters, functions or tasks of [custom_axis_infrastructure.vh](https://github.com/Xilinx/finn/pull/685/files#diff-39af3848938ed68baf3572371ee439f72025e36b143c69640dab238ab37afc29) are used where?

preusser (approved):

PR #704, opened 2022-10-26
[set_fifo_depths] Adding a new transformation to split fifos larger than the max allowed depth

hleblevec:
This adds a new transformation in `transformation/fpgadataflow/set_fifo_depths.py` called SplitLargeFifos, aiming to allow users to insert StreamingFIFO nodes with depths higher then 32768 (which is the maximum allowed by Vivado) by spliting them into smaller FIFOs.
The transform is executed during the `set_fifo_depths` step, after the configuration from the provided folding configuration file is applied (hence, in the case where the automatic FIFO sizing is not used). To control whether or not the user want to execute that transformation, a new boolean argument called split_large_fifos is added to the dataflow builder config, with a default value set to False. This allows to skip needlessly having to read through the graph to gain some compute time.


maltanar:
Couple questions and comments without having read through the PR in too much detail:
* Although creating new FIFO nodes to split up large FIFOs fits more naturally into the FINN flow (compared to a single FIFO node internally generating multiple smaller FIFO instances) this may raise some issues with regards to node naming. Normally, a 'final configuration' JSON file is generated by the flow, which contains the final sizes of all FIFOs (as well as parallelization parameters etc) and can be re-applied in the future to skip the FIFO size setting step. However, this is dependent on all nodes being named in a standardized way: the FIFO names need to be identical to what would be created by `InsertFIFO` followed by `GiveUniqueNodeNames`. If new FIFOs with new names are created after the `InsertFIFO` step, this will most likely cause problems. A better way of handling this may be to introduce FIFO splitting at a later stage after `set_fifo_depths`, e.g. as a completely new step or baked into the following steps if the option is set.
* Another use-case for FIFO splitting is resource minimization for large `impl_style=vivado` FIFOs. Example: if I need a FIFO of size 8196 this normally gets rounded-up to the nearest power-of-two depth of 16384 and wastes a lot of resources. Instead, one could split this up into two FIFOs of depth 8192 + 4. The FIFO splitting transformation introduced in this PR could be extended to support this use-case as well.

maltanar:
I made a first attempt to implement the second point here: https://github.com/Xilinx/finn/commit/bacb408a45907c145a3493f5a469a64ff9d49831

hleblevec:
Thanks @maltanar for your comments.
- Would it be fine to have it after the final configuration is exported, but still in the `set_fifo_depths` step? It would be just before `PrepareIP` is called. This way, the FIFO naming will remain the same in the config file. It would be complicated to push the transformation to another step has some of them are optional and will be skipped depending on the build configuration. 
- Thanks for this addition. Would you mind adding your commit to the PR?

maltanar:
@hleblevec I was not able to push the latest changes from `dev` into the branch on your personal fork (something to do with the latest changes in the GitHub Actions setup), so I made a new branch directly on the mainline repo: https://github.com/Xilinx/finn/tree/feature/split_large_fifos

Could you change the source branch of this PR to the new branch?

hleblevec:
@maltanar I updated the source of my fork to be the new branch you created. Tell me if this is good enough, otherwise I will have to recreate the PR because I don't seem to be able to change the source of the PR, only the target. Alternatively, if you know how to change the source, please point it out to me as I may be missing something.

PR #714, opened 2022-12-01
RTL ConvInpGen: Window parallelism

fpjentzsch:
### Overview:
Adds a new implementation style to output all window elements (pixels) simultaneously, thus enabling full SIMD (`= kernel_width * kernel_height * channels`) unfolding of the following MVAU/VVAU layer.

Previously, this was supported only for the 1D case via the HLS function `ConvolutionInputGenerator_1D_parallel`.

Includes basic readthedocs documentation.

### Depends on:

- [x] https://github.com/Xilinx/finn/pull/688 (is included)
- [x] https://github.com/Xilinx/finn/pull/723

PR #723, opened 2023-01-04
Support VVAU SIMD parallelism

fpjentzsch:
### Overview
Adds FINN compiler support for SIMD (`1 ≤ SIMD ≤ kernel_width*kernel_height`) parallelism of `Vector_Vector_Activate_Batch` HLS function.

Includes additional fixes, such as:
- BIPOLAR/TERNARY datatype support
- Correctly generated thresholds and accumulator width minimization in the unit test
- Resource estimation update (includes https://github.com/Xilinx/finn/commit/aaf03f5738d2daada44851f39e2442db1d44f9a2 from Ian Colbert)
- Remove workaround for Vivado HLS bug for bipolar threshold edge case (apparently fixed in Vitis HLS)

### Folding
SIMD parallelism is intended to work in combination with the RTL SWG (in `parallel_window` mode) to process the complete convolution window in parallel. While the HLS component supports SIMD unfolding independently from PE, the RTL SWG requires full unfolding across the channel dimension (SIMD of the SWG = PE of the VVAU) before enabling window-parallelism. Unlike the MVAU, the VVAU can't accept datawidth-converted input from a fully-parallel SWG in this case due to the depthwise data layout. As a result, the **VVAU should be unfolded by PE first, followed by SIMD**.

I'll update the `SetFolding()` transformation accordingly in the upcoming RTL SWG PR (https://github.com/Xilinx/finn/pull/714).

In the future, I could add output "re-ordering" functionality to the RTL SWG, so that it can be used in combination with a DWC to support SIMD unfolding without full PE unfolding, which can be more efficient according to my experiments.

### Depends on
- [x] https://github.com/Xilinx/finn-hlslib/pull/119 -> update commit reference accordingly


auphelia:
Thanks @fpjentzsch !

auphelia (commented):
Thanks @fpjentzsch , all looks good!

PR #724, opened 2023-01-05
[CustomOp] Fix setting of NumReps in Thresholding and N in Eltwise node

auphelia:
Fix for cppsim for Thresholding and Eltwise node.
- Thresholding: NumReps set according to mem_mode (`const`: 1 and `decoupled`: `np.prod(numInputVectors))`
- Eltwise: N set to product of output shape without channels


maltanar (changes requested):
I found a bit of time to get deeper into the numReps problem for thresholding. in short, there are two parameters in the hlslib call that causes the number of loop iterations to go up (just like the eltwise op):

template <
    unsigned **ImgDim**, unsigned NumChannels, unsigned PE,
    typename TSrcI = Identity, typename TDstI = Identity,
    int ActVal=0, typename TT, unsigned int NumSteps,
    typename TI, typename TO>
void Thresholding_Stream_Batch(hls::stream<TI> &in,
                        hls::stream<TO> &out,
                        hls::stream<ap_uint<PE*NumSteps*TT::width>> &weight,
                        int const **reps**)

on the compiler side, if we look at the attributes for the HLSCustomOp  for thresholds, it doesn't really have a concept of spatial image dimensions - it just treats the spatial image dimensions as multiple vectors to be processed because the pixel position does not matter for thresholding. so both ImgDim and numReps are derived in exactly the same way.

```
# in defines
numInputVectors = list(self.get_nodeattr("numInputVectors"))
numReps = int(np.prod(numInputVectors))
# in docompute
inp_vecs = self.get_nodeattr("numInputVectors")
total_spatial_size = int(np.prod(inp_vecs))
```

In essence, this PR solves the problem correctly by setting numReps to 1. note this is what the decoupled-mode codegen does in the docompute function as well:

```
elif mem_mode == "decoupled":
            # note that numReps is set to 1 in the invocation below, since
            # - for cppsim the repetition comes from the threshold stream reader+input
            # - for synth the unit runs continuously anyway (ap_ctrl_none)
            self.code_gen_dict["$DOCOMPUTE$"] = [
                """{}<{}, NumChannels1, PE1, {}, {}, ActVal1, ThresType1, NumSteps1>
                (in0, out, weights, 1);""".format(
                    "Thresholding_Stream_Batch",
                    total_spatial_size,
                    tmpl_args["TSrcI"],
                    tmpl_args["TDstI"],
                )
            ]
```

to summarize: there are two redundant parameter in hlslib that do the same thing, and the compiler sets both of them to the same value for const mode.
* I would suggest setting numReps=1 in a consistent way -- your PR sets it for const mode in defines while the old one for decoupled sets it in docompute . other possible improvements:
creating a define for the ImgDim template argument being passed to the function call so the generated .cpp becomes easier to read and debug, like this:
    * before Thresholding_Batch<4, NumChannels1, PE1, Slice<ap_int<16>>, Slice<ap_int<4>>>
    * after Thresholding_Batch<ImgDim, NumChannels1, PE1, Slice<ap_int<16>>, Slice<ap_int<4>>>
* eliminating one of these redundant args (ImgDim and numReps) completely from hlslib

maltanar (approved):

PR #772, opened 2023-02-22
remove unused dependency

fionnodonohoe-xlnx:
This library was added as a Brevitas dependency here: https://github.com/Xilinx/finn/commit/e73ce9aa72783daa2d4d3ed5f2b801ca7bd1a1b7

But Brevitas has since moved on and does not require this dependency anymore: https://github.com/Xilinx/brevitas/commit/75c3e1bf9e511f0a919f4d5e4dddf14fba8b53d1

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 bitstring==3.1.7
 clize==4.1.1
 dataclasses-json==0.5.7
-docrep==0.2.7
 gspread==3.6.0
 numpy==1.22.0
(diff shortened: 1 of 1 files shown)


maltanar:
Looks like Brevitas got rid of this dependency in https://github.com/Xilinx/brevitas/commit/75c3e1bf9e511f0a919f4d5e4dddf14fba8b53d1 so safe to remove from FINN too.

auphelia (approved):

PR #823, opened 2023-05-24
Hotfix: Pin IPython version

fpjentzsch:
The newest IPython version (8.13) no longer supports Python 3.8.

Solution: Pin to 8.12.2 until we switch to a newer Python version / base Docker image.


auphelia:
Thanks @fpjentzsch

PR #827, opened 2023-05-29
Adding tutorial/jupyter notebook about folding

shashwat1198:
The new notebook tutorial contains:

- Explanation about Folding in FINN (parallelization parameters PE/SIMD).
- Extend the MLP example used in the cybersecurity notebook and show how it can be optimized for latency improvement by adjusting these parameters. 
- Demonstrate trade-off for this improvement against the change in resource utilization.
- Explanation of '_StreamingDataWidthConverters_' and under what conditions are they inserted is also discussed here with an example.
- A table containing the information of how the different layers are constrained by the parallelization parameters has been included in the notebook.


auphelia (approved):
Thanks @shashwat1198 ! When the GHAs are passing for pre-commit and quicktest, this PR will be merged :)

PR #835, opened 2023-06-13
Support of transposed convolutions through pixel padding

hleblevec:
Gradually adding classes and functions to support the "pixel padding" implementation of transposed convolutions.

- [x] Custom_op for PixelPadding nodes
- [x] Test bench to validate the full deconvolution (PixelPadding +Conv) [Ongoing]
- [x] InferPixelPadding transformation to convert ONNX ConvTranspose node to FMPadding_Pixel + (optional) FMPadding_Batch + ConvInpGen + MVAU
- [x] Test bench using InferPixelPadding transformation instead of manual graph creation.


src/finn/custom_op/fpgadataflow/fmpadding_pixel.py
```diff
@@ -0,0 +1,334 @@
...
+            # stride to apply, can be non-square
+            "Stride": ("ints", True, []),
+            # number of channels in input image
+            "NumChannels": ("i", True, 0),
+            # SIMD Input parallelism
+            "SIMD": ("i", False, 1),
+            # FINN input datatype
+            "inputDataType": ("s", True, ""),
+            # shape describing input vecs per execution
+            "numInputVectors": ("i", False, 1),
```
auphelia, line 56: This node attribute is listed but never used. Usually this could be used to determine the batch size, e.g. to calculate the expected cycles. But it is not used in this node?

auphelia (commented):
Thanks @hleblevec ,
I've cleaned up a few things and adjusted the new test cases. Since the Brevitas version was changed as well, there is a few additional tests that will need to succeed before this will be merged.

src/finn/custom_op/fpgadataflow/fmpadding_pixel.py
hleblevec: This was a mistake from my side sorry, the batch size needs to be taken into account for a correct expected cycles computation

hleblevec:
Thanks @auphelia for taking the time to review this and sorry if there were things I forgot. I don't know which tests you are talking about but don't hesitate to ask if you need help running them.

PR #837, opened 2023-06-15
rename all pynq variables to zynq

fionnodonohoe-xlnx:
As more board support became available in FINN, more Zynq devices were added. As a Pynq is  Zynq device, it makes sense to abstract the variable names as such.


auphelia (changes requested):
Thanks @fionnodonohoe-xlnx for the suggestion! I think this would be a nice change but I think it is important to not break anyone's code immediately with that, so it would be important to also update the documentation, all comments and env variables.

PR #910, opened 2023-10-25
[Jenkins] Add node label to Jenkinsfile_CI to target specific machine if available

jmonks-amd:
Target finn-build machine if available, otherwise target built-in node


auphelia (approved):
Thanks @jmonks-amd !

PR #911, opened 2023-10-27
[INFRA] Exit run-docker if tool paths are invalid

jmonks-amd:
If a tool path is invalid (doesn't exist) then run-docker doesn't pass it onto docker. This can lead to errors much later in the build process. Exiting early brings the issue to the users attention immediately.

If a path doesn't exist, a warning will indicate exactly which variables are incorrect before the script exits. Here is an example of all paths badly formed:
```
VIVADO_PATH: /proj/xbuilds/SWIP/2022.2_1014_8888/installs/lin645/Vivado/2022.2 does not exist
HLS_PATH: /proj/xbuilds/SWIP/2022.2_1014_8888/installs/lin645/Vitis_HLS/2022.2 does not exist
VITIS_PATH: /proj/xbuilds/SWIP/2022.2_1014_8888/installs/lin645/Vitis/2022.2 does not exist
PLATFORM_REPO_PATHS: /opt/xilinx/platforms does not exist
```

### DIFF
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -226,15 +226,25 @@ if [ ! -z "$FINN_XILINX_PATH" ];then
   VITIS_PATH="$FINN_XILINX_PATH/Vitis/$FINN_XILINX_VERSION"
   HLS_PATH="$FINN_XILINX_PATH/Vitis_HLS/$FINN_XILINX_VERSION"
   DOCKER_EXEC+="-v $FINN_XILINX_PATH:$FINN_XILINX_PATH "
+  PATH_FLAG=0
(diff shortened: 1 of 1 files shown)


auphelia:
Hi @jmonks-amd ,

Thanks for the PR! I see your point that the warnings that the tools were not successfully mounted into the container can be overlooked, but I will not merge your PR because of the reason below.

We would like to offer users the possibility to try out certain functionality of FINN and initial investigations network implementation on our FPGAs without the full tool installations. Additionally, it allows us to quickly test certain functionality with GHA. 
So, it is acceptable and even necessary that the Docker container can be used without having the tools available.

PR #958, opened 2024-01-19
Feature/fifo sizing behaviour

jmonks-amd:
Picked from custom/NNW-final


auphelia (approved):
Thanks @jmonks-amd and @maltanar !

PR #959, opened 2024-01-19
Jenkins pipeline improvements

jmonks-amd:
Made a few changes to improve the Jenkins testing

- Moved HW testing into a separate pipeline. HW testing is environment dependent and therefore rather volatile. By splitting it, the main finn build pipeline should become more stable while HW testing can be run based on the artifacts produced by the build pipeline.
- Additional cleanup required when testing on Zynq platforms - cleanup ownership of workspace, reboot before runs.
- FINN stages are now all run in parallel to reduce overall runtime


auphelia (approved):
Thanks @jmonks-amd !

PR #963, opened 2024-01-26
Refactoring of MVAU and VVAU

mmrahorovic:
This PR refactors the MVAU/VVAU HLS custom-op in a HW custom-op, HLS custom-op and RTL custom-op.

**Depends on**
2. Support for packed MV(A)Us: PR #794.
3. Support for multi-packed DSP58s for VVUs: PR #907 

WIP


src/finn/builder/build_dataflow_steps.py
auphelia, line 1: This change looks like it has nothing to do with the new class hierarchy. Could you provide more information on this or take it out of the PR, please?

src/finn/custom_op/fpgadataflow/hlscustomop.py
auphelia, line 1: HLSCustomOp gets deprecated with the refactored system. Could you please take this out of the PR and if it is a necessary change, could you incorporate it into the appropriate new class (HWCustomOp, HLSBackend or RTLBackend)?

src/finn/custom_op/fpgadataflow/vectorvectoractivation.py
auphelia, line 1: The title of the PR indicates that it also contains a refactoring of the VVAU, for this there would be more changes required than this added function. Could you either add these changes or factor the VVAU out for now and concentrate on the changes for the MVAU?

src/finn/custom_op/fpgadataflow/hlsbackend.py
auphelia, line 1: Thanks for updating the ap int max width bound!

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -122,10 +133,14 @@ def get_nodeattr_types(self):
             # vector through the accelerator. This will get rid of any old
             # weight data from the weight FIFOs.
             "runtime_writeable_weights": ("i", False, 0, {0, 1}),
-        }
+            "preferred_impl_style" : ("s", False, "hls", {"hls", "rtl"}),
```
auphelia, line 136: This node attribute doesn't need to be added here, when executing `my_attrs.update(super().get_nodeattr_types())`, it is automatically inherited from HWCustomOp. Please remove this. In general, when reviewing the node attributes, the HW abstraction layer for the `MatrixVectorActivation` should ideally only contain node attributes that the HLS and RTL variant of the MVAU share. Please ensure that this is the case. HLS- or RTL-specific attributes can be added in the _hls or _rtl node variants, like here: https://github.com/Xilinx/finn/blob/refactor/rtl_integration/src/finn/custom_op/fpgadataflow/rtl/streamingfifo_rtl.py#L50

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -67,7 +78,7 @@ def get_nodeattr_types(self):
             "SIMD": ("i", True, 0),
             "MW": ("i", True, 0),
             "MH": ("i", True, 0),
-            "resType": ("s", False, "lut", {"auto", "lut", "dsp"}),
+            "resType": ("s", False, "dsp", {"auto", "lut", "dsp"}),
```
auphelia, line 81: I am not sure why the default is changed for this attribute? This could also be an HLS-specific node attribute.

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -385,6 +455,25 @@ def dsp_estimation(self):
         else:
             mult_dsp = 0
         return int(mult_dsp)
```
auphelia, line 457: If the resource estimation depend on the implementation style (HLS or RTL), please move the estimation functions in the respective custom ops.

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -397,6 +486,27 @@ def get_exp_cycles(self):
         exp_cycles = (mh / pe) * (mw / simd) * np.prod(num_inp_vec) / mmv
         return int(exp_cycles)
 
+# # TODO: fix exp_cycles estimations --> depends on fpga_part and clk
```
auphelia, line 489: If the function are not used (yet), could you please exclude them from this PR.

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -165,6 +180,61 @@ def infer_node_datatype(self, model):
         odt = self.get_output_datatype()
         model.set_tensor_datatype(node.output[0], odt)
 
+    def get_input_datatype(self, ind=0):
```
auphelia, line 183: There is a lot of code duplication. `get_input_datatype`, `get_weight_datatype`, `get_output_datatype`, ... are all again defined from line 510 onwards. Please remove repetitions.

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -728,6 +751,43 @@ def get_hls_compatible_threshold_tensor(self, orig_thres_matrix):
         rows between PEs is not as expected (n_thres_steps)"""
         return ret.reshape(1, pe, tmem, n_thres_steps)
 
+    def get_hls_compatible_weight_tensor(self, orig_weight_matrix):
```
auphelia, line 754: I see that you first deleted and then added this function again. For me this looks like an HLS-specific function and thus should be moved to `matrixvectoractivation_hls.py`, was there a reason to bring it back into the hw abstraction layer?

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -905,402 +965,68 @@ def generate_params(self, model, path):
...
+            mvau_thr_init = [x for x in graph.initializer if x.name == node.input[2]][0]
+            mvau_thr = np_helper.to_array(mvau_thr_init)
+            odt_is_bipolar = self.get_nodeattr("outputDataType") == DataType["BIPOLAR"]
+            out_scale = 2 if odt_is_bipolar else 1
+            out_bias = -1 if odt_is_bipolar else self.get_nodeattr("ActVal")
+            result = multithreshold(result, mvau_thr, out_scale, out_bias)
+        
+        context[node.output[0]] = result
 
     def code_generation_ipi(self):
```
auphelia, line 1031: This can be split into an HLS- and RTL-specific part and be moved into the HLS or RTL custom op files.

src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -1196,3 +1196,139 @@ def apply(self, model):
...
+                            SIMD=simd,
+                            PE=pe,
+                            inputDataType=idt.name,
+                            weightDataType=wdt.name,
+                            outputDataType=odt.name,
+                            ActVal=0,
+                            binaryXnorMode=0,
+                            noActivation=1,
+                            numInputVectors=list(mm_in_shape[:-1]),
+                            mem_mode=self.mem_mode,
```
auphelia, line 1324: As a direct result from my comments regarding the HW abstraction layer, you might need to remove some node attributes here.

tests/fpgadataflow/test_fpgadataflow_mvau.py
auphelia, line 1: For now it is fine to only test the HLS implementation of the MVAU.

auphelia (changes requested):
Thanks @mmrahorovic ! Before I merge, could you please address the comments I made and update your PR?

auphelia:
In general, this PR contains a lot of commits that are not relevant for the refactoring of the MVAU in HW abstraction layer and HLS variant. This might cause confusion and merge conflicts if you want to introduce these commits again at a later point in time.

src/finn/builder/build_dataflow_steps.py
mmrahorovic: That's correct, I'll remove it.
Nevertheless, do you think it's worth having this change upstream? My reasoning was that this particular transformation step (`step_set_fifo_depths`) should only affect the FIFOs as the name suggests. By passing an appropriate `node_filter` to the ApplyConfig transformation, we ensure that only `StreamingFIFO`-nodes are affected. This could potentially prevent confusion/bug when the folding config file has changed after the `step_apply_folding_config` step.

src/finn/custom_op/fpgadataflow/hlscustomop.py
mmrahorovic: Will do, thanks for the feedback!

src/finn/custom_op/fpgadataflow/vectorvectoractivation.py
mmrahorovic: Good point, thanks! As discussed, I'll have both the VVAU/MVAU refactoring in one PR.

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
mmrahorovic: Good point, I've removed it now. Thanks!

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
mmrahorovic: I think the name of the function is misleading. It suggests it's exclusively something that belongs to the HLS custom-op, but it's actually used in the `make_weight_file` method. That one is used both for the HLS and RTL custom-op and hence moved to the HW abstraction layer.
A more appropriate name would be: `get_hw_compatible_weight_tensor`. If you agree, I'll go on and rename that now.

mmrahorovic:
Due to many changes introduced to undo the starting point of the branch, a new branch/PR has been created to maintain a clean history. Please see: https://github.com/Xilinx/finn/pull/971

PR #966, opened 2024-01-29
[Deps] Add "pyyaml==6.0.1" dependency to Dockefile

iksnagreb:
More of a personal preference, but I assume I am not alone. I like to specify my model or "experiment" configurations as well as the metrics collected in YAML format instead of JSON, see here for an example: https://github.com/iksnagreb/attention-dummy/blob/benchmark/params.yaml

I think it does not hurt to add a YAML parser into the finn container, at least it should not interfere with anything else and I do not expect frequent changes to this dependency either.


auphelia:
Thanks @iksnagreb!

PR #967, opened 2024-01-29
Non-Interactive Docker Run

iksnagreb:
When running some automated finn builds by some experiment manager doing parameter sweeps, the jobs failed with some TTY issues from within docker, which I have tracked down to the container being connected to a non-TTY output device. So I think there should be an option to disable DOCKER_INTERACTIVE, when it is not absolutely needed. This PR is certainly not ready (it just disables DOCKER_INTERACTIVE...). This is more of reminder, to myself that I rely on this being fixed...

This needs to be properly refactored and probably be exposed to the outside so the user is in control. Maybe default to interactive runs while introducing some flag to turn this off?


iksnagreb:
This now exposes DOCKER_INTERACTIVE to the outside defaulting to "-it", i.e., the old default interactive run. Docker run can now be made non-interactive by setting DOCKER_INTERACTIVE="".

auphelia (approved):
Thanks for this, @iksnagreb. 

I looked into the quicktest CI failure I flagged, it's not a bug in the script. Because the new default is `-it`, the quicktest job (which doesn't override it) now runs with `-i`, and `-i` + `-t` on a TTY-less GHA runner triggers the input device is not a TTY. On dev this path only had `-t` (no `-i`), so Docker never checked stdin and it passed.

PR #989, opened 2024-02-21
Add C++ driver support

LinusJungemann:
Adds support for the C++ driver for FINN https://github.com/eki-project/finn-cpp-driver
The C++ driver itself is added as a submodule to FINN.

Current v1.0 driver release note: https://github.com/eki-project/finn-cpp-driver/releases/tag/v1.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ __pycache__/*
 .cache/*
 .*.swp
 *.ipynb_checkpoints*
+*.sif
 
 # Project files
 .vscode
@@ -65,6 +66,7 @@ coverage.xml
 .pytest_cache/
 
 # Build and docs folder/files
+finn_build_single_job.sh
 build/*
 dist/*
 sdist/*
--- a/.gitignore
+++ b/.gitignore
@@ -66,7 +66,6 @@ coverage.xml
 .pytest_cache/
 
 # Build and docs folder/files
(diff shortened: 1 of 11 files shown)


LinusJungemann:
Bumped driver version to v1.1: https://github.com/eki-project/finn-cpp-driver/releases/tag/v1.1

auphelia:
For reference, there should be an additional Jenkins test which tests the driver on the board, similar to: https://github.com/Xilinx/finn/blob/dev/docker/jenkins/test_bnn_hw_pytest.py

LinusJungemann:
The C++ driver depends on #1210, so that I do not have to duplicate some of the builder options. In the meantime, there are still some details that I will polish up a bit for nicer integration.

I have now moved both drivers into a single step called step_make_driver. If `pynq_driver` is in the `generate_outputs` list, then the old driver will be used. If `cpp_driver` is in the list, then the C++ driver is used. Additionally, the C++ driver version can be selected in the build config using "latest" or a commit hash.

preusser (approved):
Thanks, @LinusJungemann!
This looks good.

auphelia (approved):
Thank you, @LinusJungemann! Looks good to me as well :) I've added some small changes on code organization, tests and added the --force flag to the xclbinutil command to make it easier to re-run.

PR #1010, opened 2024-03-22
[Thresholding RTL] Prepend dummy threshold for narrow range quantization

azizb-xlnx:
binary-search thresholding RTL expects 2^N-1 thresholds, however, Brevitas implements signed activations using "narrow range" which drops one quantized values to keep the symmetry around 0. to get around this, we prepend a dummy threshold and offset the bias by -1.


auphelia (approved):
Thanks, @azizb-xlnx !

preusser (approved):

Issue #875, opened 2023-08-17
Thresholds remain too large for their DataType after RoundAndClipThresholds


hleblevec:
## Quick summary
RoundAndClipThresholds behave so thresholds values will be clipped to 1 more or 1 less than the maximum and minimum values defined by the DataType, which may lead to an error during HLS code generation.

## Details
After switching to dev branch (commit hash "fd72d48410c040ae8f715919e65ec731580235ce"), I experienced an error during the "step_hls_codegen": 

```
Traceback (most recent call last):
  File "/home/hlebleve/finn/src/finn/builder/build_dataflow.py", line 158, in build_dataflow_cfg
    model = transform_step(model, cfg)
  File "/home/hlebleve/finn/src/finn/builder/build_dataflow_steps.py", line 488, in step_hls_codegen
    model = model.transform(PrepareIP(cfg._resolve_fpga_part(), cfg._resolve_hls_clk_period()))
  File "/home/hlebleve/finn/deps/qonnx/src/qonnx/core/modelwrapper.py", line 140, in transform
    (transformed_model, model_was_changed) = transformation.apply(transformed_model)
  File "/home/hlebleve/finn/src/finn/transformation/fpgadataflow/prepare_ip.py", line 86, in apply
    _codegen_single_node(node, model, self.fpgapart, self.clk)
  File "/home/hlebleve/finn/src/finn/transformation/fpgadataflow/prepare_ip.py", line 53, in _codegen_single_node
    inst.code_generation_ipgen(model, fpgapart, clk)
  File "/home/hlebleve/finn/src/finn/custom_op/fpgadataflow/hlscustomop.py", line 288, in code_generation_ipgen
    self.generate_params(model, path)
  File "/home/hlebleve/finn/src/finn/custom_op/fpgadataflow/thresholding_batch.py", line 445, in generate_params
    self.make_weight_file(thresholds, "hls_header", weight_filename)
  File "/home/hlebleve/finn/src/finn/custom_op/fpgadataflow/thresholding_batch.py", line 350, in make_weight_file
    assert np.vectorize(tdt.allowed)(
AssertionError: Thresholds can't be expressed with type UINT4
```

[...]

PR #1030, opened 2024-04-06
Fix RoundAndClipThresholds Transformation

### LINKED ISSUE #978: Clipping thresholds to input range + 1?
I am facing the `AssertionError: Thresholds can't be expressed with type INT8` (the assertion is [here](https://github.com/Xilinx/finn/blob/f2424e7fa9bb834f192bdec632bbdebb84b64fb1/src/finn/custom_op/fpgadataflow/thresholding_batch.py#L350) in the weight file generation of `Thresholding_Batch`), which I cannot really explain, at least I cannot see how I end up with a threshold of 128 at the upper bound, except for these lines in the `RoundAndClipThresholds` transformation: https://github.com/Xilinx/finn/blob/f2424e7fa9bb834f192bdec632bbdebb84b64fb1/src/finn/transformation/streamline/round_thresholds.py#L59-L66

I wonder whether clipping thresholds to the range `(idtype.min() - 1, idtype.max() + 1)`, under the condition that at least one threshold is outside of this range, still guarantees that these thresholds are outside of the range which can be represented with `idtype`? Why is it clipped to one below/above the valid range? Is there any reasoning which I fail to see? Maybe clipping to `(idtype.min(), idtype.max())` would be correct?

iksnagreb:
See https://github.com/Xilinx/finn/issues/978 for some explanation of what it wrong with this transformation. See https://github.com/Xilinx/finn/issues/875 for more context.

- [x] Rework the `RoundAndClipThresholds` transformation to correctly respect the range of the input data type (the quantization annotation of the input) and the container data type of the thresholds tensor. Clearly defines the type-casting/type-promotion behavior to always produce integer thresholds coded as float32 initializer tensors.
- [x] Add test-cases for the `RoundAndClipThresholds` transformation, testing rounding, clipping and different data type combinations for large bit-width integers around INT24 (where the representability issues start to occur).
- [x] Validate changes against the MobileNet v1 test-case (the one failing after the initial, naive fix)
- [x] Run all quick-tests
~- [ ] Validate against the other end2end tests~
- [x] Validate against our transformer use case
~- [ ] Build finn-examples~

Closes #978


iksnagreb:
Hm, while I can successfully run the mobilenet and the cybersecurity end2end tests, the bnn-pynq tests fail immediately at the export step, this seems not to be my fault (at least it is unrelated to this PR, maybe I am missing something else or messed up some dependency?):
```
tests/end2end/test_end2end_bnn_pynq.py, line 459
      def test_export(self, topology, wbits, abits, board):
E       fixture 'topology' not found
```
Any hints on this error?

Regarding building the finn-examples: It seems these have not been updated to the recent finn release yet? Will this happen in the near future or should we skip this test for now?

auphelia:
Hi @iksnagreb,

Yes, we restructured the bnn pynq tests and they are not callable with `pytest -k bnn_pynq` anymore. You could run one of the markers, e.g `pytest -m sanity_bnn`.
If you like, I can take it from here and do the testing (apart from your transformer use case). We are currently in the process of updating the finn-examples to the new flow.

iksnagreb:
Ok, then I will skip doing the end2end and finn-examples tests now and focus on our transformer model. My dummy model already built successfully, but I would like to test the "real" model as well, just to be sure. I will also mark this as ready for review now.

auphelia (approved):

Issue #998, opened 2024-03-11
StreamingFIFO count width mismatch between Python and RTL


lecramdev:
During RTL generation in StreamingFIFO, the $COUNT_RANGE$ variable in the template for the is set to "clog2(depth-1)-1" but in Q_srl.v, the count and maxcount signals are set to "clog2(depth)". Verilator only emits a warning but during FIFO estimation with RTL simulation, the maxcount output overflows if the buffer too small. The output is then zero and the FIFO is removed from the graph. This is a problem especially in larger networks with skip connections.

auphelia:
Hi @lecramdev,

Thanks for making us aware of this! This is indeed a bug; to represent a count within 0 ≤ count ≤ depth, you actually need a counter of width `$clog2(depth+1)`. The range would then be `[$clog2(depth+1)-1:0]`.

We really appreciate contributions from the community, would you like to submit a solution for the bug to the repo? You could create a PR targeting `dev`, here are the [contribution guidelines](https://github.com/Xilinx/finn/blob/main/CONTRIBUTING.md).

PR #1044, opened 2024-04-17
Bug fix for FIFO count range mismatch between Python and RTL

auphelia:
This PR introduces a new parameter in the RTL implementation of the StreamingFIFO component which propagates the count width value from the code generation in Python to the verilog component. 
Bug fix for https://github.com/Xilinx/finn/issues/998.

### DIFF
--- a/finn-rtllib/fifo/hdl/Q_srl.v
+++ b/finn-rtllib/fifo/hdl/Q_srl.v
@@ -73,8 +73,9 @@ module Q_srl (clock, reset, i_d, i_v, i_r, o_d, o_v, o_r, count, maxcount);
 
    parameter depth = 16;   // - greatest #items in queue  (2 <= depth <= 256)
    parameter width = 16;   // - width of data (i_d, o_d)
+   parameter countwidth = $clog2(depth + 1);
 
-   parameter addrwidth = $clog2(depth);
+   localparam addrwidth = $clog2(depth);
 
    input     clock;
    input     reset;
@@ -89,10 +90,10 @@ module Q_srl (clock, reset, i_d, i_v, i_r, o_d, o_v, o_r, count, maxcount);
    input              o_r;	// - output stream ready
    wire               o_b;	// - output stream back-pressure
 
-   output [addrwidth:0] count;  // - output number of elems in queue
-   output [addrwidth:0] maxcount;  // - maximum observed count since reset
+   output [countwidth-1:0] count;  // - output number of elems in queue
+   output [countwidth-1:0] maxcount;  // - maximum observed count since reset
 
-   reg [addrwidth:0] maxcount_reg;  // - maximum count seen until now
+   reg [countwidth-1:0] maxcount_reg;  // - maximum count seen until now
    reg    [addrwidth-1:0] addr, addr_, a_;		// - SRL16 address
 							//     for data output
    reg 			  shift_en_;			// - SRL16 shift enable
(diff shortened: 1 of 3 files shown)


src/finn/custom_op/fpgadataflow/rtl/streamingfifo_rtl.py
```diff
@@ -95,12 +95,13 @@ def generate_hdl(self, model, fpgapart, clk):
         code_gen_dict["$TOP_MODULE_NAME$"] = topname
         # make instream width a multiple of 8 for axi interface
         in_width = self.get_instream_width_padded()
-        count_width = int(self.get_nodeattr("depth") - 1).bit_length()
+        count_width = int(self.get_nodeattr("depth") + 1).bit_length()
```
preusser, line 98: No more `+1` needed. `bit_length()` already performs the correct computation:
```
>>> int(16).bit_length()
5
>>> int(15).bit_length()
4
```
The former code was really only off by one.

finn-rtllib/fifo/hdl/fifo_template.v
```diff
@@ -53,7 +53,8 @@ output  $OUT_RANGE$ out_V_TDATA
 
 Q_srl #(
 .depth($DEPTH$),
-.width($WIDTH$)
+.width($WIDTH$),
+.countwidth($COUNT_WIDTH$)
```
preusser, line 57: I don't see the need for overriding this parameter. The default computation in `Q_srl` itself looks just fine to me.

finn-rtllib/fifo/hdl/Q_srl.v
```diff
@@ -73,8 +73,9 @@ module Q_srl (clock, reset, i_d, i_v, i_r, o_d, o_v, o_r, count, maxcount);
 
    parameter depth = 16;   // - greatest #items in queue  (2 <= depth <= 256)
    parameter width = 16;   // - width of data (i_d, o_d)
+   parameter countwidth = $clog2(depth + 1);
```
preusser, line 76: Is there a need to allow overriding this parameter?
Wouldn't a `localparam` (just like `addrwidth`) for an internal computation suffice?

src/finn/custom_op/fpgadataflow/rtl/streamingfifo_rtl.py
```diff
@@ -95,12 +95,13 @@ def generate_hdl(self, model, fpgapart, clk):
...
         code_gen_dict["$COUNT_RANGE$"] = "[{}:0]".format(count_width - 1)
         code_gen_dict["$IN_RANGE$"] = "[{}:0]".format(in_width - 1)
         code_gen_dict["$OUT_RANGE$"] = "[{}:0]".format(in_width - 1)
         code_gen_dict["$WIDTH$"] = str(in_width)
         code_gen_dict["$DEPTH$"] = str(self.get_nodeattr("depth"))
+        code_gen_dict["$COUNT_WIDTH$"] = count_width
```
preusser, line 104: If the `count_width` configuration is dropped above, this substitution will become obsolete.

preusser (changes requested):

src/finn/custom_op/fpgadataflow/rtl/streamingfifo_rtl.py
auphelia: Updated with latest commit to `count_width = int(self.get_nodeattr("depth")).bit_length()`

src/finn/custom_op/fpgadataflow/rtl/streamingfifo_rtl.py
auphelia: `countwidth` is not template parameter anymore, deleted from code generation.

preusser (approved):
Thanks, @auphelia!
All great.

Issue #1046, opened 2024-04-17
Read data from numpy file to hls::vector


vopade:
### New behavior
Add helper function so that data from a numpy files can be read directly into an `hls::vector` (and vice versa) instead of packing it into a large `ap_uint`.

### Motivation
Using `hls::vector` facilitates the introduction of new datatypes and also simplifies the code.

### Parts of FINN affected
`src/finn/qnn-data/cpp`

PR #1047, opened 2024-04-17
added helper function for reading from numpy files into hls::vector

vopade:
This PR adds a helper function to support reading from numpy files directly into `hls::vector`, see issue [1046](https://github.com/Xilinx/finn/issues/1046).


auphelia (approved):
Thanks @vopade!
I made some small clean ups and when the GHAs ran successfully, this PR will be merged.

PR #1087, opened 2024-05-23
Changed QONNX_TO_FINN_PYTHON verification step name to match build co…

hannahxy13:
…nfig name

QONNX_TO_FINN_PYTHON verification step is named "finn_onnx_python" in build_dataflow_config.py, but in build_data_steps.py it is named "qonnx_to_finn_python". Changed to "finn_onnx_python" to match the build config, so the verification file is created with the expected name


auphelia:
Thanks @hannahxy13 for adding the additional verification step for the builder flow!

PR #1116, opened 2024-06-29
Fix small bugs in transformations

lecramdev:


auphelia:
Thanks @lecramdev. All three fixes in this PR have since landed on dev independently:

1. DWC output width now uses the consumer's input width — src/finn/transformation/fpgadataflow/insert_dwc.py:95 (`dwc_out_width = n1.get_instream_width(in_idx)`), generalized to the resolved input index.
2. `FMPadding_rtl` is in the `simd_ops` list — src/finn/transformation/fpgadataflow/set_folding.py:139.
3. `MoveScalarLinearPastInvariants` skips fork-node producers — src/finn/transformation/streamline/reorder.py:572-582 (now with a warning + continue).

  Since all three are already covered, closing as superseded. Thanks for the fixes.

PR #1117, opened 2024-06-29
Added support for unsigned identity quantization

lecramdev:


auphelia:
Hi @lecramdev,
Thank you for the PR! 
Could you please sign-off your commit? To be compliant with our contribution guidelines: https://github.com/Xilinx/finn/blob/main/CONTRIBUTING.md
It should be possible to sign-off already pushed commits.

tests/brevitas/test_brevitas_quant_identity_export.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@pytest.mark.brevitas_export
+@pytest.mark.parametrize("abits", [2, 4, 8])
+@pytest.mark.parametrize("ishape", [(1, 15), (1, 32, 1, 1)])
+@pytest.mark.parametrize("narrow", [True, False])
+@pytest.mark.parametrize("quant", [Int8ActPerTensorFloat, Uint8ActPerTensorFloat])
+def test_brevitas_quant_identity_export(abits, ishape, narrow, quant):
+    export_path = f"test_brevitas_quant_identity_export_{abits}_{narrow}_{quant}.onnx"
+    b_act = QuantIdentity(act_quant=quant, bit_width=abits, narrow=narrow)
```
auphelia, line 53: Could you please change `narrow=narrow` to `narrow_range=narrow`? We don't get an error for this because the object can take additional kwargs but it will not be correctly derived and so `narrow` doesn't get varied in the resulting QuantNode for this test case.

auphelia (changes requested):

auphelia:
Thanks a lot for this contribution, @lecramdev! 

We've rebased your unsigned identity quantization support onto current `dev` and opened #1653 to bring it in. That PR keeps your handler change and your brevitas export test (with the `narrow_range` fix from the review here, plus an added per-channel variant and updated scratch-path handling per `tests/README.md`), so the work and credit remain yours.

Since #1653 supersedes this one, I'll close #1117 in favor of it. Thanks again!

PR #1125, opened 2024-07-06
Extend support for the resize operator

klassen9:
The PR extends/fixes the current implementations for the transformations of the resize operator in FINN. 

We wanted to streamline a model that uses a resize operator, but the current implementation only supports resize via the `scales` input. The model we used, uses the `sizes` input. `scales` and `sizes` are both responsible for determining the amount of resizing and can be translated into each other. They are also mutually exclusive. 

A problem that arose while implementing the support for `sizes` was to determine which input is used.  Different versions of the Onnx opset have a different understanding of how to mark an input as unused. As an example:

- Version 11 dictates: "If `sizes` is needed, the user must set `scales` to an empty tensor."
- Version 13 dictates: "If `sizes` is needed, the user can use an empty string as the name of `scales` in this operator’s input list."

I am thus checking if `scales` or `sizes` exist by checking if they are None or are empty to support all versions. None is returned by `model.get_initializer()` if the input is an empty string in the input list.

Furthermore, this PR adds support for the Version 10 resize operation. The Version 10 resize operation only supports the `scales` input. 

Other than the `scales` and `sizes` inputs and the input tensor, there is the `roi` (region of interest) input. The `roi` is not supported by the hardware implementation and therefore triggers an assert in `convert_to_hw_layers.py` when used.

The `test_scale_resize_nhwc.py` test was extended by a test that uses the `sizes` input instead of the `scales` input.

I also published a PR for qonnx adding the resize operation to the change_3d_tensors_to_4d transformation (https://github.com/fastmachinelearning/qonnx/pull/125).


klassen9:
Sorry, I didn't apply pre-commit correctly. Should be fixed in the second commit.

auphelia (approved):
Thanks for extending Resize support, @klassen9.
Since the PR had aged a bit against dev, I pushed a few follow-up commits directly rather than send it back:

  - Consolidated the opset/param-input detection into a shared resolve_resize_param_input helper (finn.util.basic), replacing the duplicated version-detection logic across InferUpsample, MakeScaleResizeNHWC, and
  AbsorbTransposeIntoResize (the latter previously assumed scales lived at input[2], which broke for opset-10 2-input Resize).
  - Added a guard for non-integer size/input ratios in InferUpsample, so unsupported (non-integer) upsampling fails loudly instead of silently rounding.
  - Minor cleanups: clearer naming (param_ind/param_dtype) and removed a redundant get_initializer.
  - Tests: parametrized the existing upsampler test over export opset [10, 11, 13], and added test_fpgadataflow_upsampler_sizes. A 4-input (sizes-based) Resize run end-to-end through cppsim/rtlsim.

PR #1132, opened 2024-07-15
Feature/conv ext weights

mdanilow:
- Feature input shapes of external weights layers are saved to io_shape_dict driver dictionary under "external_weights_input_shapes" key.
- These shapes are then used to recognize convolutional layers and send the parameters proper number of times in driver_base.py
- tests/end2end/test_ext_weights.py tests also the cnv-w2a2 topology along tfc-w2a2 and properly verifies the number of runtime weights files ("idmaX.npy"). This test is changed to use only hls backends for MVAU due to a different existing bug - to be addressed.
- Fixed a bug in validate.py script in the deployment package caused by the case of all wrong outputs in the test batch (np.bincount would return just number of 0's instead of number of 0's and 1's)


auphelia:
Thank you @mdanilow and @PWzor :)

PR #1151, opened 2024-08-07
[ConvolutionInputGenerator] Make infer_node_datatype update attributes

iksnagreb:
Without updating the datatype attributes of the node, there might be a mismatch between tensor annotations (the actual datatype) and the type assumed by the node. This becomes an issue for example when querying the bit-width of the stream when inserting data-width converters.


auphelia:
Thank you @iksnagreb !

PR #1157, opened 2024-08-13
[Deps] Update qonnx commit hash

lstasytis:
Bumped up qonnx to a version with the [commit](https://github.com/fastmachinelearning/qonnx/pull/127) which avoids calling mp.pool when the number of workers is set to 1, which makes debugging much easier (more details in the linked pull request).

Also updated the `auto_pad_to_explicit_padding` function path since it has changed in qonnx.


auphelia:
Thanks @lstasytis !

PR #1173, opened 2024-08-28
[Streamline] Allow AbsorbSignBiasIntoMultiThreshold to absorb any bias

iksnagreb:
Note: Keep the name indicating only handling signed-bias but adjust the datatype inference after absorbing the bias to handle signed bias as well as outputs.

Draft PR for now, related to work on streamlining for more generalized activation function support via multi-thresholds, but this could be of interest in any situation where it is easier to move some bias (not just the exported sign-bias but any scalar `Add`) upwards instead of downwards through the graph, e.g., if there is some joining `Mul` or `MatMul` operation blocking the usual flow of streamlining transformation.


auphelia (approved):
Thanks a lot for this, @iksnagreb! Nice generalization from sign-bias to arbitrary scalar bias, and it slots in cleanly. Approving. I pushed a few follow-up changes on top:
  - Renamed AbsorbSignBiasIntoMultiThreshold → AbsorbScalarBiasIntoMultiThreshold (kept a deprecated alias for backwards compatibility).
  - Added a max_bitwidth_increase guard (default 0): absorption is skipped when it would widen the output datatype beyond the tolerance, so a large positive bias no longer silently inflates the downstream datapath. Raise it to opt into the wider type.
  - Fixed the output-datatype derivation: dropped the` - narrow` term (used the true [out_bias, out_bias + num_steps] range) and replaced the abs()-based signedness pick with explicit signed/unsigned logic, which previously skipped valid straddling ranges (e.g. [-5, 10]).
  - Added a unit test covering absorb/skip cases, the tolerance knob, the signed straddling range, and the deprecation alias.

Thanks again for the contribution!

PR #1177, opened 2024-09-05
Streamlining transformations for Split and Concat nodes.

mdanilow:
New transformations in finn/transformation/streamline/reorder.py needed for yolov8 streamlining, some refactoring and generalization of existing transformations included.

- The MoveLinearPastEltwiseAdd() is now generalized by a refactored MoveIdenticalOpPastJoinOp and should be considered deprecated
- Some transformations (MoveScalarLinearPastInvariants, MakeMaxPoolNHWC, MakeScaleResizeNHWC) are fixed to check whether the node to be moved is a fork node, in which case the MoveOpPastFork would be called
 - The MoveOpPastFork() transformation is more flexible because it uses deepcopies of an original node (the get_attrs_fxn parameter is not needed anymore)
 - MoveScalarLinearPastSplit() and MoveTransposePastSplit() transformations added
 - The MoveIdenticalOpPastJoinOp() transformation is more general and supports moving arbitrary number of identical operations past the join node (instead of just two). This base class checks only the op_types of producers to verify if they are identical. Its are_producers_identical() and move_node() methods should be overwritten by a child class for specific cases of producers and join op. Implemented child classes include: MoveTransposePastJoinAdd, MoveMulPastJoinAdd, MoveAddPastJoinAdd, MoveTransposePastJoinConcat, MoveAffinePastJoinConcat some needed only are_producers_identical() method to be overwritten.
 - MoveAffinePastJoinConcat() transformation can be used for both scalar and channelwise affine operations.
 - test_move_identical_op_past_join_op.py is removed and test_move_identical_op_past_join_add.py, test_move_identical_op_past_join_concat.py, test_move_identical_op_past_split.py are added instead

### DIFF
--- a/src/finn/transformation/streamline/reorder.py
+++ b/src/finn/transformation/streamline/reorder.py
@@ -26,11 +26,10 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from copy import deepcopy
-
 import numpy as np
 import qonnx.core.data_layout as DataLayout
 import warnings
+from copy import deepcopy
 from onnx import TensorProto
 from onnx import helper as oh
 from qonnx.core.datatype import DataType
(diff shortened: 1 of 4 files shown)


src/finn/transformation/streamline/reorder.py
```diff
@@ -517,7 +518,9 @@ def apply(self, model):
 
 
 class MoveLinearPastEltwiseAdd(Transformation):
```
auphelia, line 520: If the new transformations replace this one, it can be deleted.

PR #1183, opened 2024-09-13
Make quantized activation handlers data layout aware

iksnagreb:
Cherry-picked and slightly extended commit from #901, but unrelated to all the attention stuff. Related to https://github.com/fastmachinelearning/qonnx/pull/143 and the [generalized activation function support](https://github.com/iksnagreb/activations).


tests/brevitas/test_brevitas_quant_identity_export.py
```diff
@@ -56,11 +56,22 @@ def test_brevitas_quant_identity_export(abits, ishape, narrow, quant):
...
 @pytest.mark.brevitas_export
 @pytest.mark.parametrize("abits", [2, 4, 8])
-@pytest.mark.parametrize("ishape", [(1, 15, 4, 4), (1, 32, 1, 1)])
+# The per-channel scale shape decides the channel axis (and thus the derived
+# MultiThreshold data layout), so cover channels-first and channels-last in
+# both 3D and 4D as well as the 2D-collapsed case.
+@pytest.mark.parametrize(
+    "ishape, channel_shape",
+    [
+        pytest.param((1, 8, 4, 4), (1, 8, 1, 1), id="nchw"),
```
iksnagreb, line 65: Maybe consider testing instances with broadcasting of singleton dimensions, i.e., omitting leading 1s for the channel shape? This would now be handled by `_get_channel_axis` as it seems.

I don't think there would be any reason to do the same on the input side, as thresholds/quantization never expand a new dimension.

auphelia: Thanks @iksnagreb! Good point, I've extended the quant identity and quant relu test cases accordingly.

PR #1185, opened 2024-09-18
Analytical FIFO sizing

lstasytis:
FIFO sizing is an extremely time-consuming process in terms of CPU cycles due to currently requiring RTL simulation of a model to determine the FIFO depths by tracking the behavior of the model.

Currently, FINN uses two main approaches for FIFO sizing:

The global sizing algorithm which incrementally tightens FIFO sizes while rerunning RTLSIM until a steady-state is reached of the entire model. (AutoFIFOSizingMethod.LARGEFIFO_RTLSIM)
The characteristic-function based approach which RTL simulates individual nodes and constructs characteristic functions for each one and then uses phase-shifting to determine an optimal FIFO size by finding how many cycles the input characteristic function must be shifted forward before it reaches a steady state with the output stream. (AutoFIFOSizingMethod.CHARACTERIZE)
Between these two options, the later - characteristic sizing can be dramatically sped up by computing the characteristic functions of individual nodes analytically, rather than by using RTLSIM.

This can be accomplished by manually reviewing the RTL (or HLS) code of each node and constructing a model python function which reproduces the loop behavior of only the AXI stream reads and writes, filling up the characteristic function array. 

The PR includes these python-based model functions as part of the custom_ops classes of the following nodes:

- channelwise_op
- convolutioninputgenerator
- fmpadding
- labelselect
- matrixvectoractivation
- pool
- streamingdatawidthconverter (generalized variant, very conservative estimate)
- streamingmaxpool
- thresholding
- vectorvectoractivation

Additionally, it includes modifications to /builder/build_dataflow_config.py and builder/build_dataflow_steps.py so that vivado is not called in the FIFO sizing step unless there is no characteristic function for an a node (and in that case it called to only characterize that respective node). This is achieved by introducing a new 'ipgen_ignore' node argument, which is set to true for all analytically characterized nodes once FIFO sizing is started and will force the FINN compiler to skip calling Vivado. This argument set back to false, allowing to call Vivado once the analytic FIFO sizing is finished.

Improvements to be made:

The remaining nodes in FINN should be characterized as necessary
There might exist parameter configurations for the convolutioninputgenerator and streamingmaxpool nodes where the characteristic function is inaccurate since an exhaustive search is complicated to do automatically.
Currently, the test for fifo sizing tests for exact correctness of the analytical function relative to the RTLSIM output. However, small latencies introduced at the start or end of the characteristic function by HLS do not lead to a change in final FIFO sizes. The test should be changed to compare the characteristic functions in a more relaxed manner. This would then allow to also perform an exhaustive test of all possible configurations for the nodes.
The characteristic FIFO sizing algorithm lacks support for branching networks. This is irrespective of the individual characteristic functions and should be improved in the main FIFO sizing function (transformation/derive_characteristic.py)

### DIFF
--- a/src/finn/analysis/fpgadataflow/dataflow_performance.py
+++ b/src/finn/analysis/fpgadataflow/dataflow_performance.py
@@ -29,6 +29,7 @@
 
 from qonnx.custom_op.registry import getCustomOp
 
+from finn.util.basic import decompress_string_to_numpy
 from finn.util.fpgadataflow import is_hls_node, is_rtl_node
 
 
@@ -76,4 +77,46 @@ def dataflow_performance(model):
         "critical_path_cycles": int(critical_path_cycles),
         "max_cycles": int(max_cycles),
         "max_cycles_node_name": max_node_name,
+    }
+
+
+def max_period(model):
+    """Extract maximum period among all nodes in the graph
+
+    Preconditions:
+    - model consists of HLS/RTL nodes
+    - model has cycle estimates annotated (see AnnotateCycles transformation)
+    - nodes have unique names (see GiveUniqueNodeNames)
+    - model has been characteristically derived and contains specific chr periods
+
+    Returns:
+    - max_cycles : number of cycles for slowest node
+    - max_cycles_node_name : name of slowest node
+    - critical_path_cycles : pessimistic expected latency from input to output
+    """
+    max_cycles = 0
+
+    for node in model.graph.node:
+        if node is not None and node.op_type not in [
+            "AddStreams_hls",
+            "DuplicateStreams_hls",
+            "AlignLabels_hls",
+            "StreamingFIFO_hls",
+            "StreamingFIFO_rtl",
+        ]:
+            if is_hls_node(node) or is_rtl_node(node):
+                inst = getCustomOp(node)
+                node_cycles_in = (
+                    len(decompress_string_to_numpy(inst.get_nodeattr("io_chrc_in"))[0]) // 2
+                )
+                node_cycles_out = (
+                    len(decompress_string_to_numpy(inst.get_nodeattr("io_chrc_out"))[0]) // 2
+                )
+                node_cycles = max(node_cycles_in, node_cycles_out)
+
+                if node_cycles > max_cycles:
+                    max_cycles = node_cycles
+
+    return {
+        "max_cycles": int(max_cycles),
     }
--- a/src/finn/analysis/fpgadataflow/dataflow_performance.py
+++ b/src/finn/analysis/fpgadataflow/dataflow_performance.py
@@ -33,7 +33,7 @@ from finn.util.basic import decompress_string_to_numpy
 from finn.util.fpgadataflow import is_hls_node, is_rtl_node
 
 
-def dataflow_performance(model):
+def dataflow_performance(model, use_characterized=False):
     """Extract key performance indicators from given model with dataflow nodes.
     Note that the latency (critical path) analysis is very pessimistic, it
     assumes no overlap between executions and simply sums the expected cycles
@@ -44,6 +44,15 @@ def dataflow_performance(model):
     they do not need to be specialized yet
     - model has cycle estimates annotated (see AnnotateCycles transformation)
     - nodes have unique names (see GiveUniqueNodeNames)
+
+    Parameters:
+    - use_characterized : take each node's cycle count from its token access
+      vectors (see DeriveTokenAccessVectors) instead of the cycles_estimate
+      annotation. Nodes that carry no token access vector -- the fork and join
+      nodes, whose vectors are copied from their neighbours, and FIFOs -- count
+      as zero cycles, so max_cycles is the slowest *characterized* node. Note
+      that this makes critical_path_cycles a partial sum; the characterized mode
+      exists to report max_cycles.
 
     Returns:
     - max_cycles : number of cycles for slowest node
@@ -57,7 +66,24 @@ def dataflow_performance(model):
     for node in model.graph.node:
         if is_hls_node(node) or is_rtl_node(node) or node.op_type == "Shuffle":
             inst = getCustomOp(node)
-            node_cycles = int(inst.get_nodeattr("cycles_estimate"))
+            if use_characterized:
+                # The TAVs (see DeriveTokenAccessVectors) store one entry per
+                # clock cycle for the input and the output stream back to back,
+                # so half the array length is the number of cycles the node
+                # needs for one period. A node that has not been characterized
+                # counts as 0 -- fork and join nodes and FIFOs never are, they
+                # inherit their neighbours' vectors instead of having one
+                # measured.
+                chrc_in = inst.get_nodeattr("io_chrc_in")
+                chrc_out = inst.get_nodeattr("io_chrc_out")
+                if chrc_in == "" or chrc_out == "":
+                    node_cycles = 0
+                else:
+                    cycles_in = len(decompress_string_to_numpy(chrc_in)[0]) // 2
+                    cycles_out = len(decompress_string_to_numpy(chrc_out)[0]) // 2
+                    node_cycles = max(cycles_in, cycles_out)
+            else:
+                node_cycles = int(inst.get_nodeattr("cycles_estimate"))
             if node_cycles > max_cycles:
                 max_cycles = node_cycles
                 max_node_name = node.name
@@ -77,46 +103,4 @@ def dataflow_performance(model):
         "critical_path_cycles": int(critical_path_cycles),
         "max_cycles": int(max_cycles),
         "max_cycles_node_name": max_node_name,
-    }
-
-
-def max_period(model):
-    """Extract maximum period among all nodes in the graph
-
-    Preconditions:
-    - model consists of HLS/RTL nodes
-    - model has cycle estimates annotated (see AnnotateCycles transformation)
-    - nodes have unique names (see GiveUniqueNodeNames)
-    - model has been characteristically derived and contains specific chr periods
-
-    Returns:
-    - max_cycles : number of cycles for slowest node
-    - max_cycles_node_name : name of slowest node
-    - critical_path_cycles : pessimistic expected latency from input to output
-    """
-    max_cycles = 0
-
-    for node in model.graph.node:
-        if node is not None and node.op_type not in [
-            "AddStreams_hls",
-            "DuplicateStreams_hls",
-            "AlignLabels_hls",
-            "StreamingFIFO_hls",
-            "StreamingFIFO_rtl",
-        ]:
-            if is_hls_node(node) or is_rtl_node(node):
-                inst = getCustomOp(node)
-                node_cycles_in = (
-                    len(decompress_string_to_numpy(inst.get_nodeattr("io_chrc_in"))[0]) // 2
-                )
-                node_cycles_out = (
-                    len(decompress_string_to_numpy(inst.get_nodeattr("io_chrc_out"))[0]) // 2
-                )
-                node_cycles = max(node_cycles_in, node_cycles_out)
-
-                if node_cycles > max_cycles:
-                    max_cycles = node_cycles
-
-    return {
-        "max_cycles": int(max_cycles),
     }
--- a/src/finn/builder/build_dataflow_config.py
+++ b/src/finn/builder/build_dataflow_config.py
@@ -40,9 +40,37 @@ from finn.util.basic import part_map, vitis_default_platform
 
 class AutoFIFOSizingMethod(str, Enum):
     "Select the type of automatic FIFO sizing strategy."
-
-    CHARACTERIZE = "characterize"
+    ANALYTIC = "analytical"
     LARGEFIFO_RTLSIM = "largefifo_rtlsim"
+
+
+class TAVGenerationMethod(str, Enum):
+    "Select the strategy for constructing token access vectors of an operator."
+    RTLSIM = "rtlsim"
+    TREE_MODEL = "tree_model"
+
+
+class TAVUtilizationMethod(str, Enum):
+    """Select the strategy for utilizing token access vectors of an operator
+    for buffer sizing."""
+
+    # worst-case ratio of data rates between a consumer and producer
+    CONSERVATIVE_RELAXATION = "conservative_relaxation"
+
+    # average-case ratio of data rates between a consumer and producer
+    AGGRESSIVE_RELAXATION = "aggressive_relaxation"
+
+    # compose isolated TAVs along the chain (input-arrival-constrained
+    # schedules) and size from occupancy between composed schedules; no
+    # stretch/relaxation heuristics
+    CHAIN_COMPOSED = "chain_composed"
+
+    # no relaxation, use the token access vectors as-is
+    NO_RELAXATION = "no_relaxation"
+
+    # propagate token arrival times across the whole graph, instead of comparing
+    # a producer's trace against its consumer's in isolation
+    CHAINED_TAV = "chained_tav"
 
 
 class ShellFlowType(str, Enum):
@@ -271,6 +299,27 @@ class DataflowBuildConfig:
     #: When `auto_fifo_depths = True`, select which method will be used for
     #: setting the FIFO sizes.
     auto_fifo_strategy: Optional[AutoFIFOSizingMethod] = AutoFIFOSizingMethod.LARGEFIFO_RTLSIM
+
+    #: Which strategy will be used for token access vector generation for FIFO sizing.
+    #: RTLSIM will result in performing RTLSIM for each node
+    #: to deduce the token access vectors empirically
+    #: TREE_MODEL will use the tree mode of an operator if available, avoiding the generation
+    #: of IP cores.
+    tav_generation_strategy: Optional[TAVGenerationMethod] = TAVGenerationMethod.RTLSIM
+
+    #: Which strategy will be used to turn token access vectors into FIFO depths.
+    #: See :class:`TAVUtilizationMethod` for the individual strategies; they differ
+    #: in how much slack is assumed between a producer and its consumer, and in
+    #: whether arrival times are propagated across the whole graph or compared
+    #: pairwise in isolation.
+    #: CHAINED_TAV is the default: propagating arrival times across the graph
+    #: matches the rtlsim-sized interval where the pairwise strategies do not.
+    tav_utilization_strategy: Optional[TAVUtilizationMethod] = TAVUtilizationMethod.CHAINED_TAV
+
+    #: When True, skips the resynthesis steps after fifo sizing. This makes it
+    #: possible to run the step for rapid fifo size analysis during
+    #: automatic folding optimizations or as a first approximation.
+    skip_resynth_during_fifo_sizing: Optional[bool] = False
 
     #: Memory resource type for large FIFOs
     #: Only relevant when `auto_fifo_depths = True`
--- a/src/finn/builder/build_dataflow_config.py
+++ b/src/finn/builder/build_dataflow_config.py
@@ -39,20 +39,27 @@ from finn.util.basic import hbm_boards, part_map, vitis_default_platform
 
 
 class AutoFIFOSizingMethod(str, Enum):
-    "Select the type of automatic FIFO sizing strategy."
-    ANALYTIC = "analytical"
+    """Select the type of automatic FIFO sizing strategy.
+
+    The two heuristic strategies differ only in where each node's token access
+    vector comes from. ``heuristic_analytical`` builds it from the node's tree
+    model, which needs no IP generation and no simulator; ``heuristic_rtlsim``
+    measures it per node with rtlsim, which is far slower but covers operators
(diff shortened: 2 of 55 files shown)


src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -971,3 +956,145 @@ def code_generation_ipi(self):
...
+        padding = 0
+
+        kwargs = self.prepare_kwargs_for_characteristic_fx()
+
+        # first period
+        cycles = 0
+        txn_in, cycles, counter = self.characteristic_fx_input(txn_in, cycles, counter, kwargs)
+
+        txn_in += [counter] * (period - cycles)
+        padding += period * -cycles
```
fpjentzsch, line 1066: Is this a typo, i.e. `padding += (period * 1) - cycles`?

lstasytis: Yes, will fix in the next push sometime this or next week.

fpjentzsch (changes requested):
I did not run the tests for thresholding & vvau yet.

In general we probably want to reduce the number of tests that go into the default suite. Currently the most test cases are located in

- ConvInpGen: 20.736
- Thresholding: 4.608
- VVAU: 2.304

Maybe we can still include an extended test suite via pytest.fixture or pytest_generate_tests functions @auphelia?

src/finn/custom_op/fpgadataflow/duplicatestreams.py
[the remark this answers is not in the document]
lstasytis: DeriveFIFOSizes should only get called if derive_characteristic_fxns() has been called for each node already, which are kept in the node definition classes. So I would say the optimal place would be in that same function, [like such](https://github.com/lstasytis/finn/commit/aaa498be4e5f2f3ccb38a5763b770f462b363839#diff-7409417ac962d0ff8d35509ce329586719ecc994b134e5d7793eb42d06079409L177)

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -971,3 +956,84 @@ def code_generation_ipi(self):
...
+                "in0": [0 for i in range(n_inps)],
+            },
+            "outputs": {"out": []},
+        }
+
+        mem_mode = self.get_nodeattr("mem_mode")
+        if mem_mode in ["internal_decoupled", "external"]:
+            n_weight_inps = self.calc_wmem()
+            # num_w_reps = np.prod(self.get_nodeattr("numInputVectors"))
+            io_dict["inputs"]["weights"] = [0 for i in range(1 * n_weight_inps)]
```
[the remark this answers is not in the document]
lstasytis: At some point this was causing 
> Exception: Error in simulation! Takes too long to produce output. Consider setting the LIVENESS_THRESHOLD env.var. to a larger value.
errors during rtlsim, but this got fixed somewhere along the way and I simply forgot to reintroduce it. Fixed in [this commit](https://github.com/Xilinx/finn/pull/1185/commits/aaa498be4e5f2f3ccb38a5763b770f462b363839) now.

tests/fpgadataflow/test_fifosizing.py
```diff
@@ -54,17 +54,29 @@ def fetch_test_model(topology, wbits=2, abits=2):
...
+    "method",
+    [
+        "largefifo_rtlsim_python",
+        "largefifo_rtlsim_cpp",
+        "characterize_analytic",
+        "characterize_rtl",
+    ],
 )
 @pytest.mark.parametrize("topology", ["tfc", "cnv"])
 def test_fifosizing_linear(method, topology):
```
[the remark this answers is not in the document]
lstasytis: With the [mvau fix](https://github.com/lstasytis/finn/commit/aaa498be4e5f2f3ccb38a5763b770f462b363839#diff-7409417ac962d0ff8d35509ce329586719ecc994b134e5d7793eb42d06079409L177) now in, the simulation error does not show up anymore for me, however I still get the 

> $readmem file not found 

error as well, even though the tests all pass and I manually checked that each type of fifo sizing strategy does generate the same fifo depths across the entire model. This warning is not present in the current /dev build so something is breaking, even though it is not affecting the fifo-sizing. I'm a bit stumped on what is going on here.

src/finn/custom_op/fpgadataflow/duplicatestreams.py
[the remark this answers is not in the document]
fpjentzsch: Yes, that generalization should be fine.

The situation with the `outFIFODepths` I was talking about was more a problem of my synthetic test model, which created a DuplicateStreams node and without setting `outFIFODepths` explicitly. It would still be nice if the CustomOp could set a default value depending on `NumOutputStreams` itself, but I changed my synthetic model generation for now to do the same thing the convert_to_hw transformation does (which is why this situation never failed any real FINN build flows): https://github.com/Xilinx/finn/blob/7803de53cb9585a694d254787e66de5fb916dfd7/src/finn/transformation/fpgadataflow/convert_to_hw_layers.py#L567

src/finn/transformation/fpgadataflow/derive_characteristic.py
```diff
@@ -73,114 +203,1228 @@ def apply(self, model: ModelWrapper):
...
+                                    if (prod_chrc_part >= cons_chrc_part).all():
+                                        pshift_min = pshift_cand
+                                        break
+
+                                # parent_throughput = get_parent_throughput(node, model)
+                                parent_period, producer_node = get_top_producer_period(node, model)
+                                consumer_period, consumer_node = get_top_consumer_period(
+                                    node, model
+                                )
+                                # consumer_throughput = get_consumer_throughput(cons_node, model)
```
fpjentzsch: Lots of commented out lines in this file. Should probably be removed or moved into a conditional block that is enabled by a flag if this code is still necessary for debugging or as fallback.

fpjentzsch (changes requested):

src/finn/builder/build_dataflow_config.py
```diff
@@ -278,6 +297,31 @@ class DataflowBuildConfig:
...
+    ] = TAVUtilizationMethod.CONSERVATIVE_RELAXATION
+
+    #: When True, skips the resynthesis steps after fifo sizing. This makes it
+    #: possible to run the step for rapid fifo size analysis during
+    #: automatic folding optimizations or as a first approximation.
+    skip_resynth_during_fifo_sizing: Optional[bool] = False
+
+    #: Avoid using C++ rtlsim for auto FIFO sizing and rtlsim throughput test
+    #: if set to True, always using Python instead
+    force_python_rtlsim: Optional[bool] = False
```
fpjentzsch: Wasn't this option deprecated on the dev branch?

src/finn/builder/build_dataflow_config.py
```diff
@@ -40,11 +40,39 @@
 
 class AutoFIFOSizingMethod(str, Enum):
     "Select the type of automatic FIFO sizing strategy."
-
-    CHARACTERIZE = "characterize"
+    ANALYTIC = "analytical"
     LARGEFIFO_RTLSIM = "largefifo_rtlsim"
 
 
+class TAVGenerationMethod(str, Enum):
```
auphelia: Currently users need to set both auto_fifo_strategy and tav_generation_strategy, which exposes implementation details that might be better to hide from a standard user (TAV = Token Access Vector, tree_model, etc.).
What do you think of collapsing `auto_fifo_strategy` + `tav_generation_strategy` into a single option?
```
auto_fifo_strategy:
"characterize_analytical"  # analytical sizing using tree models (fast)
"characterize_rtlsim"      # analytical sizing using rtlsim for characterization
"largefifo_rtlsim"         # legacy large-FIFO simulation approach
```

src/finn/builder/build_dataflow_config.py
```diff
@@ -40,11 +40,39 @@
...
+    "Select the strategy for constructing token access vectors of an operator."
+    RTLSIM = "rtlsim"
+    TREE_MODEL = "tree_model"
+
+
+class TAVUtilizationMethod(str, Enum):
```
auphelia: Here I am also thinking if we could come up with a naming that is for a standard user easier to understand. Even though I would see this as advanced settings, so there is a case to make that it could be less high-level. A few ideas:
```
buffer_sizing_strategy
buffer_sizing_method
sizing_from_(fifo_)characterization
depth_from_(fifo_)characterization
rate_matching_method
fifo_calculation_method
```

src/finn/analysis/fpgadataflow/dataflow_performance.py
```diff
@@ -77,3 +78,45 @@ def dataflow_performance(model):
...
+    - max_cycles_node_name : name of slowest node
+    - critical_path_cycles : pessimistic expected latency from input to output
+    """
+    max_cycles = 0
+
+    for node in model.graph.node:
+        if node is not None and node.op_type not in [
+            "AddStreams_hls",
+            "DuplicateStreams_hls",
+            "AlignLabels_hls",
```
auphelia: I believe this is not an op in finn (maybe in finn+?). Please remove from list, I believe it is in other parts of this PR as well.

src/finn/analysis/fpgadataflow/dataflow_performance.py
```diff
@@ -77,3 +78,45 @@ def dataflow_performance(model):
         "max_cycles": int(max_cycles),
         "max_cycles_node_name": max_node_name,
     }
+
+
+def max_period(model):
```
auphelia: This function feels a bit out of place as an analysis pass. Unlike dataflow_performance() which computes estimates from get_exp_cycles(), max_period() just reads back data that was already computed and stored by DeriveTokenAccessVectors.

Two options to consider:
- Move to util: Make it a utility function like `get_max_characterized_period(model)` since it's essentially a getter for TAV data (e.g. in util/fpgadataflow.py)

- Extend dataflow_performance with override behavior
```
  def dataflow_performance(model, use_characterized=False):
      if use_characterized:
          # Read from io_chrc_* attributes if available
          ...
      else:
          # Use get_exp_cycles() estimates
          ...
```
This way there's one function with a clear "use actual data if available" flag. 

  Also, the docstring appears to be copy-pasted from dataflow_performance. The "Returns" section mentions max_cycles_node_name and critical_path_cycles which aren't actually returned.

src/finn/builder/build_dataflow_steps.py
```diff
@@ -1079,8 +1183,10 @@ def step_set_fifo_depths(model: ModelWrapper, cfg: DataflowBuildConfig):
 
     # after FIFOs are ready to go, call PrepareIP and HLSSynthIP again
     # this will only run for the new nodes (e.g. FIFOs and DWCs)
-    model = model.transform(PrepareIP(cfg._resolve_fpga_part(), cfg._resolve_hls_clk_period()))
-    model = model.transform(HLSSynthIP(cfg._resolve_fpga_part()))
+    if not cfg.skip_resynth_during_fifo_sizing:
```
auphelia: Potentially redundant resynthesis, I think we can delete that part.

```
if not cfg.skip_resynth_during_fifo_sizing:
      model = model.transform(PrepareIP(...))
      model = model.transform(HLSSynthIP(...))

```
With the phase-based builder, `step_set_fifo_depths` runs in `phase_optimize_hardware`, and then `step_hw_ipgen` (which calls `HLSSynthIP`) runs later in `phase_build_hardware`. Since `HLSSynthIP` will synthesize any unsynthesized nodes including the FIFOs/DWCs created during FIFO sizing, this resynthesis step appears redundant.

Consider removing the `skip_resynth_during_fifo_sizing` config option and these lines, letting `step_hw_ipgen` handle all IP synthesis. This simplifies the config and keeps synthesis responsibility in one place (`phase_build_hardware`).

src/finn/custom_op/fpgadataflow/hwcustomop.py
auphelia: Those are not used anymore with your changes, right? If so, then get_io_chrc_in() / get_io_chrc_out() methods should also be removed and the cleanup code in build_dataflow_steps.py that deletes files from these attributes should be removed.

Also is there a cleanup that removes all the other temporary .npy files when FIFO sizing is done?

src/finn/custom_op/fpgadataflow/fmpadding.py
```diff
@@ -156,3 +157,61 @@ def execute_node(self, context, graph):
             inp_values, ((0, 0), (pad[0], pad[2]), (pad[1], pad[3]), (0, 0)), "constant"
         )
         context[node.output[0]] = np.asarray(result, dtype=np.float32).reshape(oshape)
+
+    def get_tree_model(self):
+        # key parameters
```
auphelia: Some tree model implementations have great docstrings explaining the reasoning, while others have no docstring at all.

Good examples:
  - concat.py - Explains Interval 5, staggered bursts, how it differs from passthrough assumptions
  - split.py - Explains freerunning wrapper, Interval 5, references csynth.rpt
  - lookup.py - Documents measured behavior vs initial hypothesis

Missing docstrings:
  - matrixvectoractivation.py
  - vectorvectoractivation.py
  - streamingdatawidthconverter.py
  - fmpadding.py
  - pool.py
  - labelselect.py
  - duplicatestreams.py

These just have `# key parameters` comments without explaining the rationale.

Since tree models are analytical approximations of HLS/RTL behavior, documenting why the model is structured a particular way (e.g., "measured Interval 5 from csynth.rpt" or "assumes II=1 pipelining") would help future maintainers understand and update them. Could you please add docstrings to the remaining implementations?

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
```diff
@@ -1289,3 +1313,96 @@ def code_generation_ipi(self):
             raise Exception("Unrecognized mem_mode for MatrixVectorActivation")
 
         return cmd
+
+    def get_tree_model(self):
```
auphelia, line 1435: I believe the MVAU tree model may not account for tiling and mem_mode variations.

1. Tiling (TH > 1): The RTL MVAU supports tiling which "reduces throughput by factor TH". The tree model doesn't account for this, it may produce incorrect TAVs when TH > 1.
2. mem_mode variations: The tree model doesn't differentiate between internal_embedded, internal_decoupled, external, external_mem, or dynamic modes. These have different weight fetching behaviors that could affect the read/write schedule.

Note that derive_token_access_vectors() does handle mem_mode for setting up the weight stream in io_dict, but the tree model itself (which generates the schedule) doesn't adapt its pattern.

Consider either:
  - Adding guards to return None (fall back to rtlsim) for unsupported configurations like TH > 1 or mem_mode in ["external_mem", "dynamic"]
  - Or extending the tree model to handle these cases

src/finn/custom_op/fpgadataflow/streamingdatawidthconverter.py
```diff
@@ -125,6 +126,14 @@ def get_folded_output_shape(self, ind=0):
 
         return dummy_t.shape
 
+    def get_number_input_values(self):
+        folded_ishape = self.get_folded_input_shape()
+        return np.prod(folded_ishape[:-1])
+
+    def get_number_output_values(self):
```
auphelia: Unnecessary override: get_number_output_values() is identical to the parent class implementation in HWCustomOp and can be removed.

auphelia (changes requested):
### Initial Review Feedback

Overall the structure of this PR looks excellent, I like how you split the responsibilities.

I've added inline comments throughout the code with specific suggestions and questions. The main points are around naming/API consistency, some potentially dead code (io_chrc_in_file/io_chrc_out_file), and ensuring tree models account for newer MVAU features (tiling, mem_mode variations).

One issue: I ran the MLO tests with Vivado 2024.2:
`pytest -k "finnloop and end2end" --pdb` and they're failing. Can you have a look at why that path doesn't work anymore?

I'll continue with the review tomorrow.

src/finn/custom_op/fpgadataflow/fmpadding.py
lstasytis: I actually had all of them with extensive documentation, but purposefully shrunk them down fearing it was far too many comments. If you feel it makes sense to keep 10-20line comments for each explaining the rationale, that works great, I'll reintroduce them.

auphelia: Thanks, yes I think that would make sense. Just to be able to reason about it in a few months/years.

src/finn/builder/build_dataflow_config.py
lstasytis: I'm thinking:
"heuristic_analytical"
"heuristic_rtlsim"
"largefifo_rtlsim"

src/finn/builder/build_dataflow_config.py
lstasytis: Since these are all encompassed by the heuristic strategy, I would then propose to keep the naming scheme and go with: "heuristic_fifo_sizing_method". So we have a heuristic strategy and a set of heuristic methods that are encompassed by that strategy.

auphelia: Yes, I like that

src/finn/analysis/fpgadataflow/dataflow_performance.py
lstasytis: Opted for the 2nd option, though I put the if-case deeper in the function so the code is cleaner.

Relevant dif: https://github.com/Xilinx/finn/pull/1185/changes/809ca541672f26248681e6db4e7de78357be0d73#diff-5321ca16a2880172a419b20c2031dee1b17f5eb206685b098be72ebaf91ec6baR36

src/finn/builder/build_dataflow_config.py
lstasytis: Updated with https://github.com/Xilinx/finn/commit/809ca541672f26248681e6db4e7de78357be0d73#diff-636de98950dad5dc0f496861b80651291e656a168dabd4b530d83d900daf990fL42

src/finn/builder/build_dataflow_config.py
lstasytis: Updated with https://github.com/Xilinx/finn/commit/809ca541672f26248681e6db4e7de78357be0d73#diff-636de98950dad5dc0f496861b80651291e656a168dabd4b530d83d900daf990fL42

src/finn/analysis/fpgadataflow/dataflow_performance.py
lstasytis: Artifact from the branch having at some point been merged with our branch introducing a bypass addstreams->alignLabels transformation to allow for bump-in-the-wire style deployment of finn (packet dropped capabilities), removed all references to it in 259ece7f2f4f9d2f0c77d4ae15c2f9c6cc93b000

src/finn/builder/build_dataflow_steps.py
lstasytis: Redundant indeed, removed in https://github.com/Xilinx/finn/commit/809ca541672f26248681e6db4e7de78357be0d73#diff-f13e5fdd4156183abe21debeff0069547f6325b81ce11cd43d2f99c877951f3cL1196

src/finn/custom_op/fpgadataflow/hwcustomop.py
lstasytis: Removed them, also added a cleanup function: https://github.com/Xilinx/finn/commit/809ca541672f26248681e6db4e7de78357be0d73#diff-f13e5fdd4156183abe21debeff0069547f6325b81ce11cd43d2f99c877951f3cR1012

For now it is called unconditionally after fifo sizing transformations, but, as discussed, I think it would be good to add a flag in config to skip this cleanup function so that TAVs can be reused during sizing. At that can move it to util as well.

src/finn/custom_op/fpgadataflow/fmpadding.py
lstasytis: Added the docs in ca3d1a770b663520fd2fd55e4117782724040aa7, touched up the models themselves as well in some parts, since those are being continuously refined ;)

src/finn/custom_op/fpgadataflow/matrixvectoractivation.py
lstasytis: Added tiling support, it lead to some extra tree forking inside the mvau, but that's to be expected: 
https://github.com/Xilinx/finn/pull/1185/changes/4b44947b314e07ec49502e224068f2faab1d8dc4#diff-4b91885366dc1cc0be60bc18e9801c8d71502af712cc81bb46d92041fe3dcec4L55

I set ["external_mem", "dynamic"] to default to rtlsim, since it would indeed lead to unpredictable schedules. I guess it's future work to explore how reliable a tree would be for this case (might even be device-unique).

src/finn/custom_op/fpgadataflow/streamingdatawidthconverter.py
lstasytis: Removed in https://github.com/Xilinx/finn/commit/1825a7523c3194843397b6c5f109470e4be4f87d#diff-43791f58bafc578f84a8475b89599407ae0a81f15508756040379230ec2fd8edL133

PR #1189, opened 2024-09-18
Streamlining transformations for Split and Concat nodes

mdanilow:
New transformations in finn/transformation/streamline/reorder.py needed for yolov8 streamlining, some refactoring and generalization of existing transformations included.

- The MoveLinearPastEltwiseAdd() is now generalized by a refactored MoveIdenticalOpPastJoinOp and was removed

- MoveScalarLinearPastSplit() and MoveTransposePastSplit() transformations added

- The MoveIdenticalOpPastJoinOp() transformation is more general and supports moving arbitrary number of identical operations past the join node (instead of just two). This base class checks only the op_types of producers to verify if they are identical. Its are_producers_identical() and move_node() methods should be overwritten by a child class for specific cases of producers and join op. Implemented child classes include: MoveTransposePastJoinAdd, MoveMulPastJoinAdd, MoveAddPastJoinAdd, MoveTransposePastJoinConcat, MoveAffinePastJoinConcat some needed only are_producers_identical() method to be overwritten.

- test_move_identical_op_past_join_op.py and test_linear_past_eltwise.py tests are removed and test_move_identical_op_past_join_add.py, test_move_identical_op_past_join_concat.py, test_move_identical_op_past_split.py are added instead


auphelia (approved):
Thank you @mdanilow, looks great!

PR #1193, opened 2024-09-20
Compiler integration of the Concat operation

mdanilow:
This PR depends on a PR from the finn-hlslib where the Concat operation was implemented (https://github.com/Xilinx/finn-hlslib/pull/143)

- The Concat op code is not hardcoded in the compiler anymore and it now accepts different datatypes of inputs. The output datatype is inferred.
- HLS implementation is defined for a one clock cycle, without defining NUM_REPS, so the new timeout cppsim template is also added and utilized by the operation.
- The [tests/fpgadataflow/test_fpgadataflow_concat.py](https://github.com/Xilinx/finn/compare/dev...mdanilow:finn:feature/split_and_concat?expand=1#diff-170b00f527c93b3d38b21af8562ffa7c94d142d9f65c50e333134d940545a9b3) test now tests the case of inputs being different datatypes


iksnagreb:
@mdanilow @preusser: Do you have any idea why we see IPGen failing for this (the test_fpgadataflow_concat) for 2022.2 but seemingly not for 2023.2? The error message is simply "ERROR: [HLS 200-1715] Encountered problem during source synthesis" and then it stops... I did not really debug this further and for now do not rely on this (but I planned to...), however, as we are stuck on 2022.2 at least for some of our setups, it would be nice to figure this out.

mdanilow:
@iksnagreb I have no idea xd

auphelia:
> @mdanilow @preusser: Do you have any idea why we see IPGen failing for this (the test_fpgadataflow_concat) for 2022.2 but seemingly not for 2023.2? The error message is simply "ERROR: [HLS 200-1715] Encountered problem during source synthesis" and then it stops... I did not really debug this further and for now do not rely on this (but I planned to...), however, as we are stuck on 2022.2 at least for some of our setups, it would be nice to figure this out.

This is addressed in this commit: https://github.com/Xilinx/finn-hlslib/pull/160/commits/7ab5ad9619331d7134f2b43f48c348e793ee009e

auphelia:
Thank you, @mdanilow !

PR #1194, opened 2024-09-20
Compiler integration of the Split operation

mdanilow:
This PR depends on a PR from the finn-hlslib where the Concat operation was implemented (https://github.com/Xilinx/finn-hlslib/pull/143)

- The new Split custom op is added.
- HLS implementation is defined for a one clock cycle, without defining NUM_REPS, so the new timeout cppsim template is also added and utilized by the operation.
- [tests/fpgadataflow/test_fpgadataflow_split.py](https://github.com/Xilinx/finn/compare/dev...mdanilow:finn:feature/split_and_concat?expand=1#diff-7d5b94dd239f0ac9085a02bf124de7d1fcf75b27b1cf7b6a4cbcd435e4c238d9) test added


auphelia:
Thank you @mdanilow , this looks good but there is one small thing that I would like to address before I merge. Implementing the output as an array of hls streams, causes issues with the naming of the interfaces, that is why Split_hls needs to overwrite so many functions. I will see if I can find a better solution before merging.

PR #1195, opened 2024-09-23
Different input datatypes support for AddStreams operation

mdanilow:
Different input datatypes were already supported by the HLS backend from the finn-hlslib.
This case was added to the unit test.
The unit test now also tests the InferAddStreamsLayer transform.


auphelia (approved):
Thank you, @mdanilow ! Looks great :)

PR #1202, opened 2024-09-26
New upsampler integration

mdanilow:
TODO


auphelia:
Thanks @mdanilow! This rework has since been integrated into dev in commit b0f5939c7. The current
  upsampler uses the new HI/WI/HO/WO/SIMD scheme with hls::vector streams and the upsample_nn<> hlslib function, supports rectangular upsampling, and runs against the official Xilinx/finn-hlslib (the new template landed upstream, so no fork dependency needed).

Since the functionality is already on dev, closing this as superseded. Thanks for the work!

PR #1213, opened 2024-10-11
Pyxsi enablement for alternative rtl simulation

auphelia:
This PR adds pyxsi as new backend for rtl simulation, it is replacing pyverilator in the FINN compiler and enable more simulation capabilities, like the simulation of Vivado IP.

Nota that this is the first step towards xsi integration. There will be an update in the coming weeks which replaces pyxsi with finnxsi, if you are curious, you can track the changes on https://github.com/Xilinx/finn/tree/feature/finnxsi_integration

This PR also contains changes to the RTL MVU:
It improves the usability and timing closure of the DSP-based dot-product kernels by:
- Pipelining the terminal reduction trees across parallel partial accumulations,
- Allowing to trade the available bit allocation space between weights and activations,
- Avoiding assign statements for the clock selection with difficult simulation scheduling semantics,
- Allowing very wide accumulations even spilling the high lane out of the DSP datapath, and
- Having the core DSP pipeline run freely without a high-fanout enable signal using zero-product insertion instead.


preusser (approved):
Great. Thanks!

Issue #1215, opened 2024-10-16
ERROR: [BD 5-390] IP definition not found for VLNV: xilinx.com:hls:StreamingMaxPool_hls_0:1.0


IsmailAM1999:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [ x] Test that the bug appears on the current version of the dev-branch. Make sure to include the commit hash of the commit you checked out.
- [ ] Check that the issue hasn't already been reported, by checking the currently open issues.
- [x ] If there are steps to reproduce the problem, make sure to write them down below.
- [ ] If relevant, please include the ONNX files, which were created directly before and/or after the bug.

## Quick summary
Hello,

I encountered this error in dev-branch while trying to Run step: step_set_fifo_depths [13/19]

## Details
I searched the error in vivado.log of the project, and here is the logs :

```
`ERROR: [BD 5-390] IP definition not found for VLNV: xilinx.com:hls:StreamingMaxPool_hls_0:1.0
ERROR: [Common 17-39] 'create_bd_cell' failed due to earlier errors.

    while executing
"create_bd_cell -type ip -vlnv xilinx.com:hls:StreamingMaxPool_hls_0:1.0 StreamingMaxPool_hls_0"`
```
I looked at the logs in code_gen_ipgen_StreamingMaxPool_hls_0_xpewuftl folder : 

```
WARNING: [Common 17-1221] Tcl app 'designutils' is out of date for this release. Please run tclapp::reset_tclstore and reinstall the app.
source run_ippack.tcl -notrace
INFO: calling package_hls_ip ip_types=vitis sysgen json_file=/tmp/finn_dev_i22amess/code_gen_ipgen_StreamingMaxPool_hls_0_xpewuftl/project_St>
INFO: Copied 1 ipmisc file(s) to /tmp/finn_dev_i22amess/code_gen_ipgen_StreamingMaxPool_hls_0_xpewuftl/project_StreamingMaxPool_hls_0/sol1/im>
INFO: Copied 8 verilog file(s) to /tmp/finn_dev_i22amess/code_gen_ipgen_StreamingMaxPool_hls_0_xpewuftl/project_StreamingMaxPool_hls_0/sol1/i>
INFO: Copied 8 vhdl file(s) to /tmp/finn_dev_i22amess/code_gen_ipgen_StreamingMaxPool_hls_0_xpewuftl/project_StreamingMaxPool_hls_0/sol1/impl>
[...]

PR #1222, opened 2024-11-01
[Container] Apply workaround for running multiple parallel Vivado instances

fpjentzsch:
See https://adaptivesupport.amd.com/s/article/63253?language=en_US

This issue occurs frequently for users with slow filesystems, such as for us, during the last FINN tutorial, or here: https://github.com/Xilinx/finn/issues/1215

I propose to add the workaround permanently to the Docker Container.


auphelia (approved):
Thank you @fpjentzsch, for now we can integrate this workaround.

PR #1234, opened 2024-11-22
Driving AXI bus padding to zero.

preusser:
As with the other AXI wrappers of RTL components, drive the full-byte AXI-Stream extension to zero.


auphelia (approved):
Thank you, @preusser !

PR #1251, opened 2024-12-12
Feature/fix non blocking

jsmonson:
This PR removes non-blocking assignments from the combination logic and replaces them with blocking assignments in the Q_srl.v fifo. This change is beneficial because it removes the eliminates the warnings generated by pyverilog during RTL simulation for this issue.


jsmonson:
@auphelia can you remind me which tests you would like me to run?

auphelia:
Thanks @jsmonson for creating this PR! I would recommend to run at least `test_fifosizing` and `test_fpgadataflow_fifo`, but I will do that as part of the PR review as well.

preusser (approved):
Thanks a lot, @jsmonson!

auphelia:
Thank you @jsmonson !

PR #1258, opened 2025-01-07
Fix build dir creation for tests

fpjentzsch:
Prevent annoying creation of empty temporary build directories during test discovery.


auphelia:
Thanks @fpjentzsch !

PR #1263, opened 2025-01-16
Integration of new deconvolution implementation based on the REVD2 algorithm (Ongoing)

hleblevec:
Integration of the new deconvolution implementation (see https://github.com/Xilinx/finn-hlslib/blob/feature/deconv/deconv.hpp for the HLS part).

Current known issues:

- [ ] Several failing cppsim testcases, which seem to be linked to padding when P < K-S
- [ ] rtlsim currently not working, probably due to issues in the templates
- [ ] Some testcases might fail due to loop timeout value being too small (need to identify if there are still cases where that happens)


auphelia:
Hi @hleblevec ,

Thank you for starting with this PR! I wanted to quickly point you to this new template which has an easier setup to use the time out simulation: https://github.com/Xilinx/finn/blob/feature/deconv/src/finn/custom_op/fpgadataflow/templates.py#L63

auphelia:
Thank you @hleblevec ! I will merge the PR, so that we have the latest developments on the finn repo rather than your fork.

PR #1268, opened 2025-01-28
Fix: Replace distutil copy_tree with shutil

fpjentzsch:
Replace distutil copy_tree with shutil copytree for two reasons:

1. It is deprecated
2. I ran into this obscure bug, which does not happen with copytree: https://bugs.python.org/issue22132


auphelia (approved):
Thanks @fpjentzsch , I've also replaced the occurrences in the notebooks. Not sure if needed, but you might want to pull that into the finn-plus dev branch as well.

PR #1273, opened 2025-02-05
Fix: Make Test Suite Parallelizable

fpjentzsch:
For the FINN+ CI we want to speed up the test suite.

Previously, the test suite was parallelizable only with the pytest-xdist `--dist loadfile` scheduler, because some tests assume all tests within the same file are run in-order or at least not in parallel.

This refactors the offending tests that I found (except end2end tests for now, which really depend on each other) so that we can use a better scheduler (`--dist worksteal`) for most of the test suite.


fpjentzsch:
I also made the bnn_pynq end2end tests parallelizable across different scenarios (`board, topology, wbit, abit` combination) via the `--dist loadgroup` scheduler.

auphelia (approved):
Thank you @fpjentzsch!

PR #1275, opened 2025-02-10
PR for MoveLinearPastEltwiseMul transformation

shashwat1198:
**PR Description:**

This PR introduces a new transformation, `MoveLinearPastEltwiseMul`, which optimises the computation graph by moving linear multiplication operations past elementwise multiplication operations when possible. 

- **Pattern before transformation:**  
  \[(x × A) × (y × B)\]
  
- **Pattern after transformation:**  
  \[(x × y) × (A × B)\]

Where \(x\) and \(y\) are dynamic inputs, and \(A\) and \(B\) are constant tensors. 

### Key Changes:
- The transformation identifies mul operations that can be merged and moves linear mul nodes past elementwise mul nodes in the graph.
- The transformation checks for mul operations with shared initializers and merges them, removing redundant nodes.
- The model is updated with new initializers after the transformation to ensure correctness.
- A post-transformation pass is performed to infer shapes for any updated nodes in the graph.


fpjentzsch:
We already have this in FINN+ as the ["MoveConstMulPastJoinMul" transformation](https://github.com/eki-project/finn-plus/blob/dev/src/finn/transformation/streamline/reorder.py#L2086).

Christoph (@iksnagreb) added it as part of the StreamlinePlus PR (https://github.com/eki-project/finn-plus/pull/39), which doesn't yet have an equivalent in this repository I believe.

auphelia:
> We already have this in FINN+ as the ["MoveConstMulPastJoinMul" transformation](https://github.com/eki-project/finn-plus/blob/dev/src/finn/transformation/streamline/reorder.py#L2086).
> 
> Christoph (@iksnagreb) added it as part of the StreamlinePlus PR ([eki-project#39](https://github.com/eki-project/finn-plus/pull/39)), which doesn't yet have an equivalent in this repository I believe.

Thanks @fpjentzsch, good to know it already exists in FINN+ as MoveConstMulPastJoinMul (Christoph's StreamlinePlus work, eki-project#39). Confirmed on our side though: there's no equivalent in main FINN, the MoveIdenticalOpPastJoinOp base only has Add/Concat join subclasses on dev, and no MoveConstMulPastJoinMul/MoveMulPastJoinMul has ever been part of a PR to Xilinx/finn. So rather than duplicating anything upstream, this PR is what actually brings the join-Mul case ((x*A)*(y*B) -> (x*y)*(A*B)) into main FINN.

auphelia (approved):
Thanks @shashwat1198 for bringing this upstream! 
The join-Mul case was genuinely missing from main FINN, so this fills a real gap. As @fpjentzsch noted, an equivalent exists in FINN+ as `MoveConstMulPastJoinMul` (@iksnagreb's StreamlinePlus work), but it was never part of a PR to Xilinx/finn, so this is the change that lands it here. Credit to both.

 I've pushed a small refactor on top to align it with the existing streamlining infrastructure:
- Reimplemented as `MoveMulPastJoinMul(MoveIdenticalOpPastJoinOp`), mirroring the sibling `MoveMulPastJoinAdd/MoveAddPastJoinAdd`, instead of a standalone Transformation. Constants are folded with an elementwise product, so per-channel scales are preserved (the np.sum-based Add sibling only handles scalars, this keeps your more general behavior).
- Dropped `MoveLinearPastEltwiseAdd`: its two cases are already covered on dev by MoveMulPastJoinAdd and MoveAddPastJoinAdd, so it was redundant.
- Added tests/transformation/streamline/test_move_identical_op_past_join_mul.py

LGTM to merge once CI is green.

PR #1284, opened 2025-02-25
Guard Q_srl against Instantiation with a Depth below Two

preusser:
When bigger FIFOs are split into multiple one, it can happen that FINN creates FIFOs with depth=1 which is an invalid setting.
This PR adds introduces two methods to avoid FIFOs to be created with depth=1.
- Error assertion in the RTL code when depth=1
- During FINN transformation that splits large FIFOS, when there is a FIFO to be created with depth=1, this gets increased to depth=2 instead


STFleming (approved):
Looks great. I did a test where it passed as expected and I did another test where it failed as expected.

STFleming:
image

New test case added for what we observed failing is now passing with the new fifo_depths check.

PR #1289, opened 2025-03-03
[Thresholding] Generalize data layouts for node execution

iksnagreb:
See https://github.com/fastmachinelearning/qonnx/pull/143 for the similar generalization applied to QONNX MultiThreshold


auphelia:
Does this PR depend on https://github.com/fastmachinelearning/qonnx/pull/143 ?

iksnagreb:
No, it does not depend on the QONNX PR, this mirrors the changes to the QONNX MultiThreshold node execution for the FINN custom-op Thresholding node execution. This is also still more of a workaround, we do not even have access to the layout annotations here so it is even less generalized and always guesses... Merging this should not break anything, behavior of the existing 2d and 4d layouts is the same, but as also proposed by @maltanar at the QONNX PR, we should seriously consider more robust alternatives to the data layout strings...

PR #1304, opened 2025-03-24
Prevent Vivado from losing RAM_STYLE attribute during synthesis.

preusser:
With this PR, we're preventing Vivado from losing the RAM_STYLE attribute for the RTL Thresholding component.


auphelia (approved):
Thanks @preusser !

PR #1346, opened 2025-05-28
Factored out, Consolidated AXI-lite Adapter.

preusser:
Revised and factored out implementation of AXI-lite adapter with corresponding path adjustments in the FINN compiler.


auphelia (approved):
Thanks @preusser !

PR #1364, opened 2025-06-11
Fix permission error on repeated runs with C++ driver

LinusJungemann:
Running the driver generation step, while a previous driver already exists at the destination location of the `copytree()` call in `step_make_driver` for the C++ driver, leads to permission denied errors (on some file systems such as Lustre FS).

This is the case because git sets the permission of some internal files to 444. When copying these files with `shutil.copytree`, these permissions are retained. When rerunning the driver generation, the step fails, because overwriting files with 444 permission is not possible even with `dirs_exist_ok=True` set for `shutil.copytree`.

(This bug may not affect all file systems, as some file systems do not retain permissions on file copy. This bug does affect Lustre FS.)

The fix switches `shutil.copytree` for the C++ driver generation to `shutil.copyfile`, which specifically overwrites existing files, i.e. it creates the copied files with read-write permissions and does not copy the original permissions.

### DIFF
--- a/src/finn/builder/build_dataflow_steps.py
+++ b/src/finn/builder/build_dataflow_steps.py
@@ -751,7 +751,12 @@ def step_make_driver(model: ModelWrapper, cfg: DataflowBuildConfig):
                 version=cfg.cpp_driver_version,
             )
         )
-        shutil.copytree(model.get_metadata_prop("cpp_driver_dir"), driver_dir, dirs_exist_ok=True)
+        shutil.copytree(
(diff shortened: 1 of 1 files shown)


LinusJungemann:
I have only experienced this during the driver generation (and the deployment step where the driver is copied again), but this will affect all calls where the files being copied have their permissions set to read only in the temporary FINN build directory.

auphelia (approved):
Thank you @LinusJungemann ! Let me know, if this permission error also appears at other parts of the flow, we can then change it.

PR #1365, opened 2025-06-11
New simulation backend: finnxsi

auphelia:
FINN's new rtlsim backend is finnxsi!

This PR finalizes the switch of the FINN RTL simulation backend from Verilator to Vivado simulator using [XSI](https://docs.amd.com/r/en-US/ug900-vivado-logic-simulation/Using-Xilinx-Simulator-Interface) that was started with PR https://github.com/Xilinx/finn/pull/1213. This shift enables the RTL simulation of Vivado IP, paving the way for upcoming changes, like supporting HLS floating point kernels. 

The C++ rtlsim driver and Python bindings are directly integrated into the FINN project. This integration facilitates easier maintenance and expansion while eliminating overhead associated with external dependencies.

All existing simulation functionalities have been updated to leverage the new simulation backend, including node-by-node rtlsim, stitched IP rtlsim, simulation- and characterization-based FIFO sizing and throughput testing. These advancements collectively ensure a more efficient and robust simulation environment within the FINN project.

### DIFF
--- a/finn_xsi/rtlsim_xsi.cpp
+++ b/finn_xsi/rtlsim_xsi.cpp
@@ -241,7 +241,7 @@ int main(int argc, char *argv[]) {
 			"interval_cycles\t" << max_interval << "\n"
 			"TIMEOUT\t" << (timeout > max_iters? "1" : "0") << "\n"
 			"UNFINISHED_INS\t" << itodo << "\n"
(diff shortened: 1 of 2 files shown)


STFleming (approved):
This looks great, I just had two minor things that I might have caught with error reporting.

PR #1375, opened 2025-06-25
Full zero-padding of folded wide config words.

preusser:


auphelia (approved):
Thank you @preusser, that takes care of the HLS test cases for runtime writeable thresholds.

PR #1376, opened 2025-06-25
Extended adjacency_list utility function to accept a lambda for filtering decisions.

STFleming:


auphelia (approved):
Thank you @STFleming, looks great!

PR #1380, opened 2025-06-28
Fix script to launch Singularity container

gabrielrodcanal:
Hi there,
The script `run-docker.sh` is currently broken for Singularity. The command generated contains unwanted quotes. See for example:
```
singularity exec "--env "SHELL=/bin/bash "--pwd "/lustre/home/nx08/nx08/s2081362-2/Software/finn "-B "/lustre/home/nx08/nx08/s2081362-2/Software/finn:/lustre/home/nx08/nx08/s2081362-2/Software/finn "-B "/tmp/finn_dev_s2081362-2:/tmp/finn_dev_s2081362-2 "--env "FINN_BUILD_DIR=/tmp/finn_dev_s2081362-2 " ...
```

This patch fixes the script so that FINN can be used in Singularity.

### DIFF
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -272,11 +272,12 @@ if [ -z "$FINN_SINGULARITY" ];then
 else
   SINGULARITY_BASE="singularity exec"
   # Replace command options for Singularity
-  SINGULARITY_EXEC="${DOCKER_EXEC//"-e "/"--env "}"
-  SINGULARITY_EXEC="${SINGULARITY_EXEC//"-v "/"-B "}"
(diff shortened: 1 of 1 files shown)


fpjentzsch:
When I test it on our system, I don't get these unwanted quotes with the current implementation. But your change doesn't seem to break it for me either, so I suppose we can merge.

fpjentzsch (approved):

auphelia:
Thank you @gabrielrodcanal! I will update `dev` with this commit as well and I would recommend that you work with the `dev` branch rather than with the `main` branch if you would like to make use of the latest developments.

PR #1386, opened 2025-07-06
Custom/transformer loop

jsmonson:
more updates


auphelia:
Thank you, @jsmonson !

PR #1398, opened 2025-07-25
Consolidating DSP-based MVU with Automated Flexible Laning

preusser:
Adopt automated flexible laning in DSP-based MVU implementation with all design featuring soft vectorization. This only leaves the implementation mode leveraging the INT8 mode of DSP58 explicitly diverting to an alternative kernel implementation.
Leveraging the gained flexibility, the instantiation of the DSP-based MVU in FINN is simplified by eliminating previously needed bitwidth-dependent decisions on the kernel selection.


auphelia (approved):
Thanks @preusser !

PR #1412, opened 2025-08-25
Support 64-bit buffer addresses

LeahLS:
Added 64-bit address support when setting up the DMAs in the driver template file to enable compatibility with pynq v3.1 and xrt 2.17. 

Tested running multiple finn-examples with pynq v3.1 on the AUP-ZU3 board and running the tfc-w1a1-mnist example with pynq 3.0.1 on the ZCU104 board.


auphelia:
Thank you, @LeahLS !

PR #1418, opened 2025-09-03
Enable FP32 support for thresholding

SpiritSeeker:
- Infer thresholding HW layer only when the input and thresholds are both integers or both FP32.
- Enable use of FPARG in thresholding RTL for FP32 comparisons
- Add tests for FP32 thresholding


src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -176,15 +176,21 @@ def apply(self, model):
                 thl_thres_shape = model.get_tensor_shape(thl_threshold)
                 idt = model.get_tensor_datatype(thl_input)
                 tdt = model.get_tensor_datatype(thl_threshold)
-                # skip conversion for layers with float input
-                if not idt.is_integer():
+
+                # only infer layers where input and thresholds are
```
auphelia, line 180: I think the flow could also be that the input is integer and the threshold values are still float. The transformation `RoundAndClipThresholds` can be called on MultiThreshold or Thresholding layers. (https://github.com/Xilinx/finn/blob/dev/src/finn/transformation/streamline/round_thresholds.py#L51)
Could you relax this constraint for the conversion? I think the check if the thresholds are int when the inputs are int, should be done in the code generation instead and if float thresholds are detected, the user can be asked to run `RoundAndClipThresholds` before code generation.

auphelia:
Thank you @SpiritSeeker ! Looks great :)

fpjentzsch:
@SpiritSeeker @auphelia On our end this broke 3 tests. Please double check and see if you need these fixes on your side as well.

**1) test_convert_to_hw_layers_cnv_w1a1[False]** 

The input quantizer is now unexpectedly converted to thresholding_hls. I fixed this here: https://github.com/eki-project/finn-plus/pull/95/commits/9d4c6540639d63057bc9746844e3749700e3ee46

Note that 1) an additional `AbsorbConsecutiveTransposes()` was needed and 2) `InferConvInpGen()` had to be moved to not leave 2x Im2Col in front of the pooling layers (this adjustment was probably missed during the recent consolidation of the pooling operators).

I wonder if this will have negative/unexpected implications for other models used in examples/regression tests...

**2) test_fifosizing_linear**

Error encountered in minimize_accumulator_width:
> Exception: Could not find a suitable int datatype for -3.4028234663852886e+38

I fixed this by enforcing FP32 datatype for thresholds in case of FP32 inputs: https://github.com/eki-project/finn-plus/pull/95/commits/1e99b397e94df51fcc67cca360e49ee3570dd02d

**3) test_split_large_fifos**

As with the cnv, the model now has 1 additional thresholding layer + DWC, leading to 2 additional FIFOs that were not accounted for. Fixed here: https://github.com/eki-project/finn-plus/pull/95/commits/d72df88494512be63f63ae7a4dd36f7a921367db

fpjentzsch:
For some models, like the KWS example, this also uncovers streamlining problems that we should probably address with proper layout conversion in the front-end and more robust handling of Transpose/Reshape/Flatten operators. See https://github.com/eki-project/finn-plus/issues/101

PR #1439, opened 2025-09-26
Softmax

STFleming:
This PR introduces a HLS Softmax layer into FINN.

Specifically it contains:
* `HWSoftmax` - a hardware abstraction layer.
* `HWSoftmax_hls` - for instantiating the HLS softmax IP
* `InferHWSoftmax` transformations for inferring the `HWSoftmax` layers.
* Tests for testing the softmax implementation

The HLS implementation of the softmax unit lives in finn-hlslib and it is structured as a three-stage pipeline with the following stages:

1. The first stage performs a pass over the input vector to determine the maximum value.
2. The second stage normalises the input vector to between (0,1] by subtracting the max and applying the exponential.
3. The last stage performs an elementwise division of the exponential values with the sum of the exponentials.

A template parameter specifies the input datatype; however, the output datatype is fp32. Internally, a fixed point datatype is used after the first stage as the dynamic range of the input floats after this normalisation is well defined. To avoid any loss of accuracy, the fixed point value maintains 23 fractional bits.


fetch-repos.sh
```diff
@@ -32,7 +32,7 @@ QONNX_COMMIT="d72fb8cf1608e87b07c28ff165943016f78dcc38"
 FINN_EXP_COMMIT="0724be21111a21f0d81a072fccc1c446e053f851"
 BREVITAS_COMMIT="b106358c4169d8a9b68cb2a531aa795417d74887"
 CNPY_COMMIT="8c82362372ce600bbd1cf11d64661ab69d38d7de"
-HLSLIB_COMMIT="a19482ba6886f6f26aff11b10126a82ce0dd7ab1"
+HLSLIB_COMMIT="1dd4743436ca0346ab54dd88edb456678a9eee25"
```
auphelia: Currently this PR points to a finn-hlslib branch, when this PR (https://github.com/Xilinx/finn-hlslib/pull/164) is merged, then the commit hash can be updated and this PR be merged.

auphelia: It now points to finn-hlslib dev, which contains: https://github.com/Xilinx/finn-hlslib/pull/165

PR #1458, opened 2025-10-16
Increase rtl and fifo-depth sim timeouts (fixes #1457)

### LINKED ISSUE #1457: RTL and fifo-depth sim timeouts
When building the upscaled/unfolded bnn-pynq finn-examples for the AUP-ZU3 board, rtl and fifo-depth simulation timeouts occur.

LeahLS:
Slightly increased the rtl and fifo-depth simulation timeouts to fix issue #1457 to not run into simulation timeouts when building the upscaled bnn-pynq finn-examples for the AUP-ZU3 board.


fpjentzsch:
@auphelia Okay, although I think we should ideally detect mismatches between actual and estimated latency/interval via some kind of is-close comparison in the PyTests or in separate regression tests, not via detecting a simulation timeout.

PR #1475, opened 2025-11-06
Fixing X_INTERFACE syntax.

preusser:
Quick fix in `X_INTERFACE_PARAMETER` attributes.


STFleming (approved):
Looks good to me. both the FIFO tests and fmpadding tests are passing.

PR #1509, opened 2026-01-20
Optimizing Index Computations in SWG for Improved Timing

preusser:
RTL optimizations for the Sliding Window Generator (SWG) default template to improve FPGA fabric mapping and timing.

Changes
- Register pre-computation of control signals: Added registered ReadingLast and ReadingDone signals to reduce combinational logic depth
- Register position tracking: Added PositionZero register to avoid comparison in critical path
- Improved fabric mapping: Switched MUX and ADD operations order for better LUT utilization
- Refactored position computation: Reformulated counter update logic using bitwise operations instead of comparisons (e.g., (~counter & (MAX-1)) == 0 instead of counter == MAX-1)
- Fixed operator precedence issue in counter reset logic
- Reset element counter to -1 fix for edge case handling

Motivation

These changes reduce the critical path through the SWG control logic by moving comparisons into registers and restructuring arithmetic operations for more efficient FPGA mapping.


STFleming (approved):
Looks good to me and tests are passing on my end.

fpjentzsch (approved):
LGTM, all tests passing on our end.

I did a small spot check on 16 exemplary SWG configurations and the average Fmax actually decreased from 194 MHz to 192 MHz. Maybe I didn't cover any cases where these changes would have a positive impact on timing.

Where there any specific edge cases that caused you trouble before?

PR #1515, opened 2026-01-29
Untie Vivado optimization for baked-in weights.

preusser:
Don't insist on the chosen `RAM_STYLE` if weights are baked in as this prevents absorbing them into the downstream logic.


STFleming (approved):
Looks good to me!

Issue #1517, opened 2026-02-03
minimize_accumulator_width() can fail due to floating point rounding errors


fpjentzsch:
Hi,

we believe the range arithmetic in [minimize_accumulator_width()](https://github.com/Xilinx/finn/blob/37f368527a4d07dea92008f577708bf5dea629f4/src/finn/custom_op/fpgadataflow/thresholding.py#L124) of the thresholding operator (and potentially other operators) is problematic because integer values are processed as (single-precision) floating point, which can lead to rounding errors (e.g., resulting threshold bitwidth of 32 instead of 33) on certain systems.

See https://github.com/eki-project/finn-plus/issues/153

PR #1524, opened 2026-02-14
Fixed processing of integers for minimize_accumulator_width()

rothej:
minimize_accumulator_width() had an issue where threshold and input range values were processed as float32, which led to rounding errors - mixed numpy float32 scalars with python int for math operations. Fix was to convert values to python int before performing arithmetic operations.

Test function was added to check for correct behavior at datatype boundaries for int16, int24, and int32. 

Fixes issue #1517


rothej:
Short term fix done! @auphelia Let me know if I should make any adjustments to these

PR #1556, opened 2026-04-07
CI: Improve Docker reliability, test parallelism, multi-agent image sharing

merkelmarrow:
This PR sets up the groundwork for fast CI for FINN, while delivering some initial speedups. The idea is to make the infrastructure more friendly to test parallelisation across multiple machines. Some additional reliability changes have been included.

In future, this foundation can be used to split test stages and achieve much faster CI times across the build fleet.

### docker/Dockerfile.finn

Add `apt-get update` before XRT .deb install to prevent stale apt cache 404s when fetching XRT dependencies. This issue has caused run failures. The fix is a documented Docker best practice - [Docker Best Practices](https://docs.docker.com/build/building/best-practices/) _"Always combine RUN apt-get update with apt-get install in the same RUN statement."_

### docker/jenkins/Jenkinsfile

- Replace `docker system prune -a -f` with targeted `container prune -f` + `image prune -f` (preserves build cache).
- Add a "Build Docker Image" stage that pre-builds the image before parallel test stages. When `FINN_DOCKER_SHARED_DIR` is set, exports the image to shared NFS. Otherwise, behaviour stays identical.
- Add `pytest-xdist` parallelisation (`-n $NUM_PYTEST_WORKERS --dist worksteal`) to the Sanity Unit Tests stage.

### run-docker.sh

Add optional multi-agent Docker sharing. When `FINN_DOCKER_SHARED_DIR` is set and the image isn't available locally, the script will load it from shared storage. Degrades gracefully on failure by falling back to a local build. Uses `flock` to prevent concurrent loads on the same host and `docker tag` aliasing to handle `git describe` hash length differences across machines.

## How it works

The image sharing is controlled by two environment variables set in the Jenkins job DSL:

- `FINN_DOCKER_PREBUILT` tells stages to skip rebuilding (the only exception being the "Build" stage itself)
- `FINN_DOCKER_SHARED_DIR` sets where to export/load the image.

The two flags are independent: FINN_DOCKER_SHARED_DIR alone works correctly (agents just run a redundant cache-hit build after loading). FINN_DOCKER_PREBUILT=1 alone also works correctly (agents skip building and use whatever image is already local). Used together they provide the best performance by building once and distributing.

## Testing

- Full FINN run completed (run 319), most stages showed a decrease in wall clock time.
- Partial FINN run completed (run 323), verified the same logic but centralised in `run-docker.sh` for maintainability.
- Verified locally, no behaviour regressions when `FINN_DOCKER_SHARED_DIR` is unset.

### DIFF
--- a/docker/Dockerfile.finn
+++ b/docker/Dockerfile.finn
@@ -84,6 +84,7 @@ RUN if [ -z "$LOCAL_XRT" ] && [ -z "$SKIP_XRT" ];then \
 COPY requirements.txt $XRT_DEB_VERSION.* /tmp/
 
 RUN if [ -z "$SKIP_XRT" ];then \
+    apt-get update && \
     apt install -y /tmp/$XRT_DEB_VERSION.deb && \
     rm /tmp/$XRT_DEB_VERSION.deb; fi
 
--- a/docker/jenkins/Jenkinsfile
+++ b/docker/jenkins/Jenkinsfile
@@ -1,19 +1,27 @@
 pipeline {
   agent none
+  options {
+    timeout(time: 72, unit: 'HOURS')
+  }
   parameters {
     booleanParam(name: 'fpgadataflow', defaultValue: false, description: 'Run fpgadataflow tests')
     booleanParam(name: 'sanity', defaultValue: true, description: 'Run sanity hardware and unit tests')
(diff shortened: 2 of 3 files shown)


auphelia (approved):
Thank you @merkelmarrow, looks good to me!

PR #1557, opened 2026-04-10
Requant layer

auphelia:
This PR introduces a new Requant hardware custom operation that provides an optimized alternative to MultiThreshold for uniform threshold patterns. The requant operation computes:

`output = clip(round(input * scale + bias), min, max)`

This is functionally equivalent to MultiThreshold when thresholds are uniformly spaced, but more efficient as it replaces threshold comparisons with a single multiply-add operation.

New Components:
  - Requant base HWCustomOp with scale/bias parameters
  - Requant_rtl backend for integer inputs using DSP-based compute
  - Requant_hls backend for float inputs using HLS math functions
  - RTL implementation supporting DSP48E1 (7-series), DSP48E2 (UltraScale+), and DSP58 (Versal)
  - InferRequantLayer transformation to convert MultiThreshold and Quant nodes

Deprecated/Removed:
  - Float2Int operation (replaced by Requant)


preusser-amd (approved):
Looks good to me, @auphelia!
The doubled wrapping is unfortunate but appears unavoidable to me as we need to make the SystemVerilog type `shortreal` available for parameter substitution.

PR #1567, opened 2026-04-21
Pr/extract multithreshold transform

sgerber-amd:
## ExtractMultiThresholdScaleBias Transform
Adds the ExtractMultiThresholdScaleBias transform to extract `out_scale` and `out_bias` from MultiThreshold nodes into seperate Mul/Add operations, resetting attributes to 1.0 and 0.0 respectively. 
**Purpose:**  RTL threshold layers (`InferThresholdingLayer`) require `out_scale=1.0` and `out_bias=0.0` to operate correctly. This transform prepares MultiThreshold nodes for RTL specialization.
**Integration:** Applied conditionally in streamline step when `cfg.standalone_thresholds = True` (RTL path only). No-op for HLS flows.
**Testing:** Unit test validates the extraction logic.


src/finn/builder/build_dataflow_steps.py
auphelia, line 1: I wouldn’t integrate `ExtractMultiThresholdScaleBias` into the default builder flow for now (i.e., in build_dataflow_steps.py). I think it’s a good optimization that should be available as an option, but I’m not convinced it should always be used. In some scenarios it could introduce additional Add and Mul nodes that would then need to be streamlined again. Running it after streamlining could therefore undo streamlining efforts.
I would suggest adding a hint in the error message that was raised for you, mentioning that this transformation exists and that the user could try it (for example, as a custom step). What do you think?

auphelia (changes requested):

src/finn/builder/build_dataflow_steps.py
auphelia: Thanks for addressing this!

auphelia (approved):
Looks good to me. Thanks @sgerber-main!

PR #1573, opened 2026-04-24
Add PWPolyF piecewise polynomial activation layer

ollycassidy13:
- Add PWPolyF, an RTL activation layer that approximates GELU, SiLU, Sigmoid,                                           
  and Tanh using piecewise degree-2 polynomials evaluated via Horner's method                                           
  on cascaded DSPFP32 FMA units. Two DSPs per PE, no BRAM, single-cycle                                                 
  throughput (II=1).                                                                                                    
- Coefficients and per-function clamping config are delivered through a                                                 
  SystemVerilog package (`pwpolyf_pkg.sv`) containing a `func_cfg_t` struct,                                            
  regenerated at build time for the node's K value by `generate_coeffs_pkg()`.                                          
- `InferPWPolyFLayer` converts both the explicit `PWPolyF` custom op and                                                
  standard ONNX activations (`Gelu`, `Sigmoid`, `Tanh`, `Sigmoid+Mul` for                                               
  SiLU, `Div+Erf+Add+Mul+Mul` for GELU with dynamo=True / opset < 20) to                                                
  PWPolyF HW layers with default K=3.                                                                                   
- PE folding handled by `SetFolding`. Resource estimates: 2 DSPs + ~200 LUTs                                            
  per PE, 0 BRAM.


tests/fpgadataflow/test_fpgadataflow_pwpolyf.py
STFleming, line 1: Nice work thanks @ollycassidy13! 

This looks like a good test that covers a lot however I think there are some key areas that might be missing currently. It looks like the current test does a good job of testing the node inference and compiler level transformations, however it seems to be missing code generation tests and RTL simulation tests. 

For the PR these should be included and I would probably suggest at minimum:
1. RTL `node-by-node` and `stitched_ip` tests (see [here](https://github.com/Xilinx/finn/blob/beb7c7be798183679546f49b657c6a059a98247b/tests/fpgadataflow/test_fpgadataflow_layernorm.py#L141) for an example on how to do this). We probably want some varying of PE and SIMD here also. 
2.  `generate_hdl` smoketests to make sure that the compiler is generating code appropriately. 
3.  smoketests for generating the coefficients packages. 

I would also probably include some tests from the RTLSimulation to make sure that the latency estimates are accurate. A good example can be found [here](https://github.com/Xilinx/finn/blob/beb7c7be798183679546f49b657c6a059a98247b/tests/fpgadataflow/test_fpgadataflow_shuffle.py#L473)

src/finn/custom_op/fpgadataflow/pwpolyf.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def bram_estimation(self):
+        # coefficients stored in LUT ROM, not BRAM
+        return 0
+
+    def uram_estimation(self):
+        return 0
+
+    def dsp_estimation(self, fpgapart=None):
+        # two DSPFP32 FMA instances per PE (Horner evaluation)
```
STFleming: This is assuming a DEGREE = 2 instance, what happens when we have DEGREE > 2?

src/finn/custom_op/fpgadataflow/pwpolyf.py
```diff
@@ -0,0 +1,188 @@
...
+    def get_number_output_values(self):
+        return np.prod(self.get_folded_output_shape()[:-1])
+
+    def get_exp_cycles(self):
+        # II=1, latency amortised over stream length
+        return np.prod(self.get_folded_output_shape()[:-1])
+
+    def lut_estimation(self):
+        pe = self.get_nodeattr("PE")
+        return 200 * pe
```
STFleming: It might be good to have a StitchedIP test to see how accurate this is. It also looks like `DEGREE` is not being considered here but I imagine LUT usage will likely scale with degree.

STFleming (changes requested):
Nice work @ollycassidy13 this is a very useful PR!

My biggest current concern is testing (see comments) it would be great if we could work on improving the test coverage together to verify thing more robustly. In particular making sure that codegen and the RTL is verified. Please ping me if you have any questions. 

Thanks again for the great work!

docs/finn/pwpolyf.md
auphelia, line 1: Could you please move this into the components subsection? https://github.com/Xilinx/finn/tree/dev/docs/finn/components

src/finn/custom_op/fpgadataflow/rtl/pwpolyf_rtl.py
```diff
@@ -0,0 +1,196 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
auphelia: You can use the SPDX-identifier for the copyright header instead and the year is not necessary anymore.

```
```

src/finn/custom_op/fpgadataflow/rtl/pwpolyf_rtl.py
```diff
@@ -0,0 +1,196 @@
...
+import os
+import shutil
+import struct
+
+from finn.custom_op.fpgadataflow.pwpolyf import PWPolyF
+from finn.custom_op.fpgadataflow.rtlbackend import RTLBackend
+from finn.util.pwpolyf import CLAMP_CFG, NUM_OCTAVES, SUPPORTED_FUNCS, _fit_coefficients
+
+
+def _float_to_hex(f):
```
auphelia: FINN already has float-to-hex packing in finn.util.data_packing.array2hexstring (line 92 handles floats). Please use:

`array2hexstring(np.array([f]), DataType["FLOAT32"], 32, prefix="").upper()`

This keeps it consistent with other RTL backends. Alternatively, BitArray(float=f, length=32).hex.upper() would also work.

src/finn/custom_op/fpgadataflow/rtl/pwpolyf_rtl.py
```diff
@@ -0,0 +1,196 @@
...
+    """Convert a Python float to a 32-bit IEEE 754 hex string."""
+    return "%08X" % struct.unpack("!I", struct.pack("!f", float(f)))[0]
+
+
+def generate_coeffs_pkg(K, degree=2, num_samples=1000):
```
auphelia: `generate_coeffs_pkg` could be a class method of PWPolyF_rtl. Other RTL backends use instance methods like `prepare_codegen_rtl_values` or `generate_params` that access node attributes via self. This would be more consistent and avoid passing K and degree explicitly:

```
  def _generate_coeffs_pkg(self, num_samples=1000):
      K = self.get_nodeattr("K")
      degree = self.get_nodeattr("degree")
      ...
```

src/finn/custom_op/fpgadataflow/pwpolyf.py
```diff
@@ -0,0 +1,207 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
auphelia: You can use the SPDX-identifier for the copyright header instead and the year is not necessary anymore.

```
```

src/finn/custom_op/fpgadataflow/pwpolyf.py
```diff
@@ -0,0 +1,207 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import math
+import numpy as np
+from qonnx.core.datatype import DataType
+
+from finn.custom_op.fpgadataflow.hwcustomop import HWCustomOp
+
+# Piecewise polynomial constants matching the RTL module
+_NUM_OCTAVES = 5
```
auphelia: Could you add a comment explaining why NUM_OCTAVES is a fixed constant rather than a configurable node attribute like K and degree?

src/finn/custom_op/fpgadataflow/pwpolyf.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def get_num_segments(self):
+        K = self.get_nodeattr("K")
+        return 1 + 2 * _NUM_OCTAVES * (1 << K)
+
+    def make_shape_compatible_op(self, model):
+        oshape = self.get_normal_output_shape()
+        return super().make_const_shape_op(oshape)
+
+    def infer_node_datatype(self, model):
```
auphelia, line 61: Since PWPolyF only supports FLOAT32 (as checked in verify_node), we could warn or raise an error if the incoming datatype is not FLOAT32, rather than silently updating the attribute to an unsupported type. Additionally, outputDataType is read but never updated. For a FLOAT32→FLOAT32 op it should be set to match the input.

```
def infer_node_datatype(self, model):
   node = self.onnx_node
   idt = model.get_tensor_datatype(node.input[0])
   assert idt == DataType["FLOAT32"], (
       f"{node.name}: PWPolyF requires FLOAT32 input, got {idt}"
   )
   self.set_nodeattr("inputDataType", idt.name)
   self.set_nodeattr("outputDataType", idt.name)
   model.set_tensor_datatype(node.output[0], idt)
```

Alternatively, since input and output are always FLOAT32, you could simplify to a single `dataType` node attribute instead of separate `inputDataType`/`outputDataType`.

src/finn/custom_op/fpgadataflow/pwpolyf.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def get_normal_input_shape(self, ind=0):
+        nch = self.get_nodeattr("NumChannels")
+        vecs = list(self.get_nodeattr("numInputVectors"))
+        return tuple(vecs + [nch])
+
+    def get_normal_output_shape(self, ind=0):
+        return self.get_normal_input_shape()
+
+    def get_number_output_values(self):
```
auphelia: `get_number_output_values`: This is identical to the base HWCustomOp implementation. Can be removed.

src/finn/util/pwpolyf.py
```diff
@@ -0,0 +1,237 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
auphelia: You can use the SPDX-identifier for the copyright header instead and the year is not necessary anymore.

```
```

tests/fpgadataflow/test_fpgadataflow_pwpolyf.py
```diff
@@ -0,0 +1,753 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
auphelia: You can use the SPDX-identifier for the copyright header instead and the year is not necessary anymore.

```
```

tests/fpgadataflow/test_fpgadataflow_pwpolyf.py
```diff
@@ -0,0 +1,753 @@
...
+from finn.transformation.fpgadataflow.create_stitched_ip import CreateStitchedIP
+from finn.transformation.fpgadataflow.hlssynth_ip import HLSSynthIP
+from finn.transformation.fpgadataflow.prepare_ip import PrepareIP
+from finn.transformation.fpgadataflow.prepare_rtlsim import PrepareRTLSim
+from finn.transformation.fpgadataflow.set_exec_mode import SetExecMode
+from finn.transformation.fpgadataflow.set_fifo_depths import InsertAndSetFIFODepths
+from finn.transformation.fpgadataflow.specialize_layers import SpecializeLayers
+from finn.util.pwpolyf import PiecewisePolyActivation
+
+test_fpga_part = "xcve2002-sbva484-2MP-e-S"
```
auphelia: The FPGA part used in the tests is not supported in Vivado 2022.2. Please use a part that works with both Vivado 2022.2 and 2024.2, such as the VCK190 part (xcvc1902-vsva2197-2MP-e-S).

auphelia (changes requested):

STFleming (approved):

auphelia (approved):
LGTM!

Thanks @ollycassidy13 and @STFleming for this excellent contribution adding PWPolyF support for piecewise polynomial activations on Versal!

I've made some cleanup changes on top of your work:

Code Reorganization:
- Created finn/custom_op/general/pwpolyfunction.py as the ONNX custom op with constants, coefficient fitting, and numpy-based execution
- Created finn/custom_op/general/__init__.py with QONNX registry registration
- Refactored finn/util/torch_hw_modules.py to only contain the PyTorch module, importing constants from the new package
- This separates concerns: ONNX graph execution vs PyTorch training/export


Code Cleanup:
- Extracted shared _pwpolyf_eval() helper in torch_hw_modules.py
- Merged _raise_pwpolyf_unsupported() into _pwpolyf_rtl_possible() in specialize_layers
- Fixed make_shape_compatible_op to use Identity node (handles unknown input shapes)

Tests:
- Consolidated tests into fewer, parameterized tests following FINN conventions
- Changed export test to use Brevitas export_qonnx instead of torch.onnx.export
- Added test_brevitas_pwpolyf.py for mixed Brevitas + PWPolyF model export


Documentation:
- Shortened pwpolyf.rst to focus on essentials (usage, supported patterns)

PR #1575, opened 2026-04-30
AddCLSToken

ollycassidy13:
Prepends a learned class token, supplied through cls_data, to each input sequence of patch tokens. The class token and patch tokens are transferred as SIMD-wide folds of ELEM_WIDTH-bit elements.


src/finn/custom_op/fpgadataflow/rtlbackend.py
```diff
@@ -93,6 +93,9 @@ def execute_node(self, context, graph):
             node = self.onnx_node
             inputs = {}
             for i, inp in enumerate(node.input):
+                nbits = self.get_instream_width(i)
+                if nbits == 0:
```
STFleming: Shouldn't a 0-width input throw an assertion of some kind? Is there a use case where a 0-width input is needed?

tests/fpgadataflow/test_fpgadataflow_addclstoken.py
```diff
@@ -0,0 +1,294 @@
...
+    return model, cls_values
+
+
+@pytest.mark.fpgadataflow
+def test_convert_concat_to_addclstoken():
+    model, cls_values = _make_concat_model()
+    patches = np.arange(12, dtype=np.float32).reshape(1, 3, 4)
+    expected = np.concatenate([cls_values, patches], axis=1)
+
+    ret = execute_onnx(model, {"patches": patches})
```
STFleming: We've been caught out a few time with tests that rely on particular input names. It might be better to extract the name to use from the model itself `{model.graph.input[0].name : patches}` sometimes FINN transformations can rename this inputs and it silently fails.

src/finn/transformation/fpgadataflow/specialize_layers.py
```diff
@@ -389,6 +389,7 @@ def apply(self, model):
                 node.input,
                 node.output,
                 domain="finn.custom_op.fpgadataflow." + impl_style,
+                name=node.name,
```
STFleming: @auphelia is this okay? Or is it breaking any naming assumptions downstream?

STFleming: This should be redundant https://github.com/Xilinx/finn/blob/a15e2ac7f98a4491417b157aabd5021f78a63db9/src/finn/builder/build_dataflow_steps.py#L633

STFleming (changes requested):
Thanks Olly looks good! Just some small questions / suggestions.

PR #1579, opened 2026-05-06
ONNX Where Operator

ollycassidy13:
Adds a streaming FINN hardware implementation of the ONNX Where(cond, X, Y) operator, selecting X where cond is true and Y otherwise, with full NumPy-style multidirectional broadcasting across the three inputs.
  - Where HWCustomOp and Where_rtl RTL backend, with shape/datatype inference, resource estimation, and cppsim/rtlsim execution.
  - finn-rtllib/where: where.sv ternary-select core plus three input_gen.sv replay buffers that broadcast-expand each operand to the output shape; Verilog/SV wrapper templates.
  - InferWhereLayer transform to lower standard ONNX Where to the HW op, with broadcast-compatibility, shape, and datatype checks.


src/finn/custom_op/fpgadataflow/rtl/where_rtl.py
```diff
@@ -0,0 +1,159 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
STFleming: Please replace the license header with 
```
```

src/finn/custom_op/fpgadataflow/rtl/where_rtl.py
```diff
@@ -0,0 +1,159 @@
...
+from finn.custom_op.fpgadataflow.rtlbackend import RTLBackend
+from finn.custom_op.fpgadataflow.where import Where
+
+
+def _rtlsrc_dir():
+    return os.environ["FINN_ROOT"] + "/finn-rtllib/where/hdl"
+
+
+class Where_rtl(Where, RTLBackend):
+    """RTL implementation of Where."""
```
STFleming: Maybe add a little more info here, or say the ONNX operator `Where` maybe?

src/finn/custom_op/fpgadataflow/rtl/hwwhere_rtl.py
```diff
@@ -0,0 +1,159 @@
...
+        cond_rtl_shape = self._rtl_shape(cond_shape)
+        x_rtl_shape = self._rtl_shape(x_shape)
+        y_rtl_shape = self._rtl_shape(y_shape)
+        assert out_rtl_shape[-1] % pe == 0, "PE must divide the output innermost dimension"
+
+        rtlsrc = _rtlsrc_dir()
+        template_path = rtlsrc + "/where_template.v"
+        with open(template_path, "r") as f:
+            template = f.read()
+        core_template_path = rtlsrc + "/where_core_template.sv"
```
STFleming, line 56: I'm not quite sure why we have two templates here. Usually what happens is we have an `<op name>_axi.sv` and a `<op name>_template.v` and inside the `<op name>_axi` module is instantiated with parameters being passed through. 

Maybe there is a good reason for having this additional template complexity logic, but it's not instantly obvious to me. 

Check out FMPadding RTL as an example: 
[template](https://github.com/Xilinx/finn/blob/dev/finn-rtllib/fmpadding/hdl/fmpadding_template.v) 
[axi](https://github.com/Xilinx/finn/blob/dev/finn-rtllib/fmpadding/hdl/fmpadding_axi.sv)

STFleming: Ahh is this because you want to do packed array literals and you need to parameterise? Maybe make this clear in a comment.

src/finn/custom_op/fpgadataflow/where.py
```diff
@@ -0,0 +1,238 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
STFleming: Please replace the lic header with the newer version

```
```

src/finn/custom_op/fpgadataflow/where.py
```diff
@@ -0,0 +1,238 @@
...
+                "inputDataType": ("s", True, ""),
+                "outputDataType": ("s", False, ""),
+                "ram_style": (
+                    "s",
+                    False,
+                    "auto",
+                    {"auto", "block", "distributed", "ultra"},
+                ),
+                "inFIFODepths": ("ints", False, [2, 2, 2]),
+                "outFIFODepths": ("ints", False, [2]),
```
STFleming: I think `outFIFODepths` is the same as the base class so could be dropped from here. I might add a comment to `inFIFODepths` for the reason why it is being overridden.

src/finn/custom_op/fpgadataflow/where.py
```diff
@@ -0,0 +1,238 @@
...
+        attr_odt = self.get_nodeattr("outputDataType")
+        if attr_odt != "" and DataType[attr_odt] != idt:
+            warnings.warn(
+                "outputDataType changing for %s: %s -> %s"
+                % (node.name, str(DataType[attr_odt]), str(idt))
+            )
+        self.set_nodeattr("outputDataType", idt.name)
+        model.set_tensor_datatype(node.output[0], idt)
+
+    def verify_node(self):
```
STFleming: Can remove and just use the base one unless specific checks are being inserted here. It feels like a lot of the assertions elsewhere will catch most issues.

src/finn/custom_op/fpgadataflow/hwwhere.py
```diff
@@ -0,0 +1,238 @@
...
+        cond = context[node.input[0]]
+        xval = context[node.input[1]]
+        yval = context[node.input[2]]
+
+        result = np.where(cond.astype(bool), xval, yval)
+        context[node.output[0]] = np.asarray(result, dtype=np.float32).reshape(
+            self.get_normal_output_shape()
+        )
+
+    def bram_estimation(self):
```
STFleming, line 291: Is this true? I thought with the `input_gen` we would potentially be getting some BRAMs here.

src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -1334,6 +1334,98 @@ def apply(self, model):
...
+class InferWhereLayer(Transformation):
+    """Convert ONNX Where(condition, X, Y) into a streaming Where layer."""
+
+    def apply(self, model):
+        graph = model.graph
+        node_ind = 0
+        graph_modified = False
+        for node in graph.node:
+            node_ind += 1
+            if node.op_type != "Where" or node.domain not in ["", "ai.onnx"]:
```
STFleming: Does `ai.onnx` appear in practice, I thought the empty domain would be enough?

@auphelia, is it an issue that the fpgadataflow op has the same name as the ONNX level op?

src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -1334,6 +1334,98 @@ def apply(self, model):
...
+            out_dt = model.get_tensor_datatype(node.output[0])
+            if out_dt is not None and out_dt != x_dt:
+                continue
+
+            cond_dt = model.get_tensor_datatype(cond_name)
+            if cond_dt is None:
+                model.set_tensor_datatype(cond_name, DataType["BINARY"])
+                cond_dt = DataType["BINARY"]
+            if cond_dt != DataType["BINARY"]:
+                continue
```
STFleming, line 1421: For some of these continue breaks I feel like we should add some Warnings for why the infer was aborted. Like what we have for some of the other layers [example](https://github.com/Xilinx/finn/blob/9e45c313062e7cba64d804aee025bb2b1ea99ef9/src/finn/transformation/fpgadataflow/convert_to_hw_layers.py#L1236)

tests/fpgadataflow/test_fpgadataflow_where.py
```diff
@@ -0,0 +1,615 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
STFleming: Please use the new lic header 

```
```

tests/fpgadataflow/test_fpgadataflow_where.py
STFleming, line 1: A general comment on the testing would be to utilise test parameterisation more rather than having duplicated code and avoiding "same body, different shape tuple" code duplication.

tests/fpgadataflow/test_fpgadataflow_where.py
```diff
@@ -0,0 +1,615 @@
...
+def test_where_stitched_ip_synth_ooc():
+    model = _prepare_where_stitched_ip_model(pe=2)
+    model = model.transform(CreateStitchedIP(FPGA_PART, CLK_NS, run_pnr=True))
+    ret = parse_ooc_synth_results(model.get_metadata_prop("vivado_stitch_proj"))
+    assert ret is not None
+
+    assert ret["LUT"] > 0
+    assert ret["FF"] > 0
+    assert ret["DSP"] == 0
+    assert ret.get("BRAM_18K", 0) == 0
```
STFleming: Isn't this a function of how big the problem size is and we might start seeing BRAMs being used for larger shapes, or will the `input_gen.sv` always use LUTRAM in these cases? It looks like it's RAM_STYLE is being set to `auto` so I would expect that this might happen.

We might want to improve this test and test against the estimates resource usage for particular problem sizes. @auphelia or might doing this too much slow down testing time?

STFleming (changes requested):
Thanks Olly, this is a great PR! 

I have a few comments / suggestions.

STFleming (approved):
Looking really nice! Thanks @ollycassidy13!

STFleming:
Also just a note that all tests were passing on my side for this.

auphelia (approved):
Thanks a lot for this contribution, @ollycassidy13! Nice work.

A few changes pushed on top during review:

  - **`Where` → `HWWhere` / `Where_rtl` → `HWWhere_rtl`**: the op reused the standard ONNX `Where` op_type verbatim (only HW op to do so). Renamed following the `HWSoftmax` precedent, which also let us drop the special node-`domain` guard in the convert-to-hw transform.
  - **Estimator BRAM fix**: `_input_gen_buffer_specs` replayed (coef 0) whenever an operand dim was 1, but `where.sv` only replays when the operand dim  is 1 *and* the output dim is > 1. On leading size-1 (batch) dims this over-estimated buffer depth (8192 vs the RTL's 4) and inflated BRAM. Mirrored the RTL condition; updated the resource test and reworked the OOC synth test to use a genuine broadcast.
  - **FLOAT16 rtlsim**: widened the `RTLBackend.execute_node` datatype assert to accept `float16`; added INT4/FLOAT16 to the execution test.
  - **Cleanups**: inlined the auto-BRAM `1024`-bit threshold (FINN style) and tightened the `_input_gen_buffer_specs` docstring to note it mirrors `input_gen.sv`.

I am waiting for a few more tests to finish, but then it's good to go. Thanks again!

PR #1593, opened 2026-05-27
Report Invalid Custom Op Configs

ollycassidy13:
Every exception around getCustomOp(...) and inst.set_nodeattr(...) gets caught in the same section. This means two very different cases look the same:

  1. A normal ONNX node has no FINN custom-op wrapper. That should be skipped.
  2. A FINN custom-op exists, but the config is wrong, for example a misspelled attribute name. That should fail loudly.

The old except Exception: pass hid both. So a bad config could appear to apply successfully while silently leaving the model misconfigured.


auphelia (approved):
Looks great, @ollycassidy13 ! Thanks for this contribution.

PR #1595, opened 2026-06-03
Harden build dir deletion against transient NFS failures [CI 1/9]

merkelmarrow:
This is PR 1 of 9 of a series intended to make CI faster and more robust.

This patch adds `robust_rmtree`, a helper that replaces `shutil.rmtree` in several cleanup and test sites. `robust_rmtree` simply retries a build tree deletion if it fails due to `ENOTEMPTY` and `EBUSY` errors associated with Network File System storage. These errors are a race that can happen when a tool subprocess closes a file just prior to removing the tree and caused non-deterministic test errors in a small percentage of samples.

This change reflects the need to move our testing suite to NFS to make better use of distributed testing infrastructure.

`make_build_dir()` was also adjusted to stop leaving orphan directories in /tmp.

### DIFF
--- a/src/finn/transformation/fpgadataflow/cleanup.py
+++ b/src/finn/transformation/fpgadataflow/cleanup.py
@@ -29,9 +29,9 @@
 
 import os
 import qonnx.custom_op.registry as registry
-import shutil
 from qonnx.transformation.base import Transformation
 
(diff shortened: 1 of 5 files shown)


auphelia:
Thank you, @merkelmarrow !

PR #1596, opened 2026-06-03
Close the build log handler before returning [CI 4/9]

merkelmarrow:
This is PR 4 of 9 of a series intended to make CI faster and more robust.

`build_dataflow_cfg` added a FileHandler to the run log but never removed or closed it, so the open descriptor kept the log file and caused rmtree issues later on. Extract the step loop into a helper and run it inside a try/finally that removes and closes the handler on every exit path.

The loop is in a helper so the try/finally can be added without indenting the whole loop.


auphelia:
Thank you @merkelmarrow ! Looks good to me.

PR #1597, opened 2026-06-04
Retry dependency fetches and fail fast in entrypoint [CI 6/9]

merkelmarrow:
This is PR 6 of 9 of a series intended to make CI faster and more robust.

This patch hardens several observed failure modes in FINN CI. The idea is to fail early when the necessary dependencies could not be fetched, and retry on transient network errors that would otherwise break a long test pipeline. These transient errors are made more frequent when CI is sharded and several shards are fetching dependencies simultaneously (as they will be later in this PR series).

### Changes

- add set -e fail-fast to both the dependency fetch script and the container entrypoint
- added a retry helper with exponential backoff around network steps.
- replace git pull with `git fetch --tags --force` followed by `git checkout <pinned-commit>` which is safer for detached-HEAD, necessary for stricter `set -e` policy.
- detect an unusable clone by testing for a resolvable HEAD instead of just the existence of the directory (valuable in a retry case)
- fail loudly when a dependency can't be checked out at its pinned commit.
- download board-file archives with `wget -qO <file>` so a retried download overwrites a partial file instead of writing "pynq-z1.zip.1" which the `unzip` would ignore.
- wrap the qonnx pyproject.toml workaround in an EXIT trap that restores the file even if the editable install fails, so pip failure cannot leave the mounted qonnx checkout half-renamed (necessary with `set -e` policy on pip installs).
- quote shell variables as a drive-by improvement (make commands robust to spaces, empty values, and globbing)

### Testing

Full functional Jenkins validation runs, no transient network-error build failures since deployed on live CI (20+ full runs).

### DIFF
--- a/docker/finn_entrypoint.sh
+++ b/docker/finn_entrypoint.sh
@@ -28,6 +28,9 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
+# Fail fast so a partial deps/ tree is caught early
+set -e
+
 export HOME=/tmp/home_dir
 export SHELL=/bin/bash
 export LANG="en_US.UTF-8"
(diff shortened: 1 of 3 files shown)


auphelia:
Thanks @merkelmarrow!

PR #1600, opened 2026-06-05
Route generated-script tool calls through a resolver [CI 2/9]

merkelmarrow:
This is PR 2 of 9 of a series intended to make CI faster and more robust.

This patch is the first building block of a new testing infrastructure model. Most of FINN's tests complete relatively quickly on a single machine. The slow parts are almost exclusively the calls to xilinx tools such as Vivado, v++, and vitis-run/hls. This PR introduces a tool resolver hook in basic.py that allows a full testing suite to be centred on a single machine, while intercepting heavy tool calls and delegating to a compute cluster through user-written shims.

I chose this approach over a pure PATH shim entirely on the deploy-side because in practice it introduced obscure problems and unintended shell/subprocess interactions and is less user friendly for those wanting to set up their own testing. This model is explicit for the end user who wants to run their own infrastructure:

1. Write shim scripts that delegate the calls for the explicitly specified tools (vivado, v++, vitis_hls, vitis-run) to your compute cluster (e.g. IBM LSF).
2. Arrange them flat in a directory and point FINN_TOOL_DIR_OVERRIDE to the directory.
3. FINN should now delegate heavy processes.

Behaviour is mostly identical to before if FINN_TOOL_DIR_OVERRIDE env var isn't set aside from clear fail-fast error if tool is missing.

#### Xelab note

Xelab is added to the resolver in a separate PR as it requires special handling.

#### Maintenance note

New heavy tool calls should be added to and called through the resolver. Every time a tool is added (which should be quite rare), it will fail CI tests that use the tool until a corresponding shim is written on the deployment side (assuming FINN_TOOL_DIR_OVERRIDE is set on that deployment). This is the consequence of an explicit model and preferable to a silent fallback that would eventually cause OOMs or load saturation on the single test machine. The tool calls now provide a clear fail-fast error if they are missing.

### DIFF
--- a/src/finn/custom_op/fpgadataflow/rtl/finn_loop.py
+++ b/src/finn/custom_op/fpgadataflow/rtl/finn_loop.py
@@ -45,7 +45,7 @@ from finn.custom_op.fpgadataflow import templates
 from finn.custom_op.fpgadataflow.hwcustomop import HWCustomOp
 from finn.custom_op.fpgadataflow.rtlbackend import RTLBackend
 from finn.transformation.fpgadataflow.annotate_cycles import AnnotateCycles
-from finn.util.basic import make_build_dir
+from finn.util.basic import make_build_dir, resolve_xilinx_tool
 from finn.util.create import adjacency_list
 from finn.util.data_packing import npy_to_rtlsim_input, rtlsim_output_to_npy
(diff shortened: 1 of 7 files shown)


merkelmarrow:
Changed to stop using tmp_path

auphelia:
Looks good, thank you @merkelmarrow !

PR #1601, opened 2026-06-08
Finn adv

surajkarki66:


auphelia:
Hi @surajkarki66, I believe your PR only seemed that large because you tried to target finn `main`, try targeting `dev` instead. This is where we usually submit PRs to. What is your contribution about?

PR #1605, opened 2026-06-12
Generalise rtlsim package ordering and route xelab via resolver [CI 3/9]

merkelmarrow:
This is PR 3 of 9 of a series intended to make CI faster and more robust.

This patch hardens the FINN XSI rtlsim compile path and completes the xelab handling through the resolver (introduced in #1600). 

xelab aborts if a SystemVerilog package is analysed after a module that imports it, so the compile step floats packages ahead of their importers. To date its been done with a hardcoded whitelist which needs to grow reactively. ca70a7876 ("Sort list of hdl sources to make sure configs come first") floated `swg_pkg`, then 240dd536c ("Add mvu_pkg to be loaded first") added `mvu_pkg`. Each new package is an expensive red CI run and a manual whitelist add, which doesn't scale well as finn-rtllib grows. I've replaced the whitelist with a simple heuristic (*_pkg.{sv,v}). While doing so, the xelab call is now routed through the tool resolver so it can be delegated like the other heavy tools, and a xelab failure now fails fast with the real error instead of a missing output `.so` file.

### Changes

- float any `*_pkg.{sv,v}` source ahead of regular modules by basename, instead of matching a fixed set of names. The ordering is a stable partition, so the caller's relative order within each group is preserved.
- moved the ordering helpers into a small finn_xsi/srcutil.py that doesn't import the `xsi` C-extension so they are unit-testable without finn-xsi built.
- route the xelab call through resolve_xilinx_tool so it can be delegated to HPC/LSF clusters
- add a "check" flag to `launch_process_helper` so a non-zero exit raises immediately. Previously a failed xelab returned quietly and surfaced later with a FileNotFoundError on the missing `.so` file.
- unit tests for reordering

### Testing

Reproduced the failure directly with xelab, a package listed after its importer fails with `VRFC 10-2989`, while floating it first fixes it. Confirmed that a realistic new source will break the old whitelist method. Many functional rtlsim validation runs in Jenkins.

### DIFF
--- a/finn_xsi/finn_xsi/adapter.py
+++ b/finn_xsi/finn_xsi/adapter.py
@@ -13,9 +13,10 @@ import os
 import os.path
 import re
 from finn_xsi.sim_engine import SimEngine
+from finn_xsi.srcutil import order_pkg_first
 from typing import Optional
 
-from finn.util.basic import launch_process_helper
+from finn.util.basic import launch_process_helper, resolve_xilinx_tool
 
 
 def locate_glbl() -> Optional[str]:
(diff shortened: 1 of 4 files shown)


auphelia (approved):
Thanks a lot for the contribution @merkelmarrow ! This is a really nice set of improvements overall.
Also appreciate you making the source file ordering more future-proof (hopefully) by generalizing the *_pkg handling instead of relying on a fixed name list.

PR #1606, opened 2026-06-17
Tiled MVAU assertion 

STFleming:
The tiled MVU (TH>1) groups input vectors into tiles of TH and assumes the tile size divides the number of input vectors evenly. A partial final tile (numInputVectors % TH != 0) never completes in weights_buff_tile, deadlocking the dataflow. Fail loudly at codegen rather than producing a silently hanging design.


auphelia (approved):
Looks good to me, thanks @STFleming !

PR #1566, opened 2026-04-21
Feature/tiling mlo

d-kor:
Summary
Adds tiled matrix-vector activation support (TH > 1) for Versal DSP58, together with the MLO / FINNLoop external-memory weight-fetch path, and integrates both into the standard MVAU/FINNLoop test flows.

Key changes
  - Tiled MVU RTL (finn-rtllib/mvu_tiled/): new cu_mvau_tiled, input_gen, acc_stage, weights_buff_tile, mvu_tiled_axi (+ wrapper and testbench). Replaces the earlier replay/transpose approach with an input_gen-based datapath.
  - MLO weight fetch (finn-rtllib/fetch_weights/): reworked fetch_weights + local_weight_buffer, moved out of finn-rtllib/mlo/. External-memory weights packed per byte-aligned IWSIMD group using the memstream PE-flipped / TH-untiled ordering; per-SIMD byte
  alignment fixed for stitched-IP rtlsim.
  - MVAU custom op (matrixvectoractivation.py, matrixvectoractivation_rtl.py, hwcustomop.py): TH-aware folding, memstream depth scaling with TH, DSP58 double-pumped compute wiring, and HLS interface updates. Obsolete MMAU/ram_p_c files removed (tiled MMAU RTL kept under mvu_tiled/ for future use).
  - Loop rolling / FIFO sizing: MLO external_mem is now set by loop rolling rather than the user; FINNLoop m_axi weight ports associated by MVAU name; set_fifo_depths updates for the loop/MLO flow.
  - Tests: tiled MatMul + MLO variants folded into test_fpgadataflow_mvau.py, test_fpgadataflow_finnloop.py, and test_loop_rolling.py; guards added for unsupported configs (tiling + double-pumping disallowed; SIMD==1 + pumpedCompute disallowed).

  Notes
  - Tiling (TH > 1) and clock double-pumping are mutually exclusive by design and skipped in tests.
  - Depends on the XSI double-pumped stimulus-timing fix (separate PR) for the pumpedCompute rtlsim path.

PR #1607, opened 2026-06-23
Add DDR-based MLO support for Zynq devices

klassen9:
Extends FINN's MLO flow (loop-based weight offloading via FINNLoop) to store weights and intermediate frames in DDR rather than only HBM, enabling MLO on Zynq-class boards without HBM. Memory placement is selectable via build config, with address offsets generated automatically and threaded through the RTL, Zynq build, and PYNQ driver. The existing HBM path is unchanged by default.

• Build config: new mlo_weight_mem option to select between HBM and DDR
• Address handling: new AssignMemoryOffset transform assigns offsets to weight and intermediate-frame regions in a shared DDR layout. Offset and base-address signals added to fetch_weights/loop_control, plus a new address_config RTL module.
• Zynq build & driver: make_zynq_proj/stitched IP wired up for DDR base addresses. PYNQ driver extended for MLO.
• RTL fixes: byte-aligned intermediate_frames padding, mux prioritizes intermediate frames to fix a batch-size bug, corrected fetch-weight sizing and .dat generation.
• Testing: New end-to-end DDR MLO test plus extended FINNLoop tests covering additional MLO edge cases.

### TODO:

- [x] Rename m_axi_hbm (no longer HBM-specific now that DDR is supported).
- [x] Merge with dev after PR #1566 is merged.
- [x] Some test cases currently fail due to the fetch weight component. At least some of these are fixed by #1566.
- [x] Look into tiled MatMul + MLO issues

### Misc:

-  Parallel write_bitstream jobs via NUM_DEFAULT_WORKERS
-  verify_step now has a batched argument which allows for real batching


preusser-amd (approved):
Thanks so much, @klassen9!

src/finn/builder/build_dataflow_config.py
```diff
@@ -382,6 +382,9 @@ class DataflowBuildConfig:
     #: loop_body_hierarchy and loop_body_range
     mlo: Optional[bool] = False
 
+    # DDR or HBM
+    mlo_weight_mem: Optional[str] = "HBM"
```
auphelia: A few observations and suggestions regarding the `mlo_weight_mem` config parameter:

1. Add MLO guard to `step_assign_ddr_weight_offsets`: The step should check both `cfg.mlo` and `cfg.mlo_weight_mem` together:
```
  def step_assign_ddr_weight_offsets(model, cfg):
      if cfg.mlo and cfg.mlo_weight_mem == "DDR":
          model = model.transform(AssignMemoryOffset())
      else:
          print("Skipping step_assign_ddr_weight_offsets (MLO not enabled or not using DDR)")
      return model
```
2. Consider changing default to DDR: Currently defaults to HBM, but there's no system deployment flow for HBM in finn dev yet (only rtlsim works). Since DDR deployment is functional, it might be safer to default to DDR until HBM deployment is ready.
3. Board-based auto-detection: Could we infer the memory type from the board instead of requiring manual config? For example:
    - Zynq boards (ZU3, KV260, etc.) → always DDR
    - V80, U50, U280, U55c → HBM available
    - U200, U250 → DDR
(This logic already exists in alveo_build.py for external weights. We could add a helper like `board_has_hbm(board_name)` and use it to auto-select.)

4. If we are leaving the builder config parameter instead of auto-detection, consider adding a check in build_dataflow_checks.py to warn/error if:
    - `mlo=True and mlo_weight_mem="HBM"` but board doesn't have HBM
    - `mlo=True and mlo_weight_mem="HBM"` and bitfile generation requested (since HBM deployment isn't ready)

What do you think about these suggestions?

auphelia: I actually thought about this a bit more and what do you think of turning this into a FINNLoop node attribute:

Instead of the mlo_weight_mem build config parameter, consider making mem_type a node attribute on FINNLoop:

1. Add `"mem_type": ("s", False, "DDR")` to `FINNLoop.get_nodeattr_types()` in finn_loop.py
2. Remove `mlo_weight_mem` from DataflowBuildConfig
3. Update `AssignMemoryOffset` to check `loop_inst.get_nodeattr("mem_type") == "DDR"` 
4. `step_assign_ddr_weight_offsets` can always run (transformation checks `mem_type` internally) or is guarded by `cfg.mlo`

Users can set mem_type via their folding config:

```
  {
      "FINNLoop_0": {"mem_type": "DDR"}
  }
```

tests/fpgadataflow/test_fpgadataflow_finnloop.py
```diff
@@ -751,3 +768,196 @@ def test_finnloop_end2end_mlo(
...
+    cfg = build_cfg.DataflowBuildConfig(
+        output_dir=tmp_output_dir,
+        steps=steps,
+        synth_clk_period_ns=10.0,
+        board="AUP-ZU3_8GB",
+        shell_flow_type=build_cfg.ShellFlowType.VIVADO_ZYNQ,
+        rtlsim_batch_size=100,
+        standalone_thresholds=True,
+        mlo=True,
+        mlo_weight_mem="DDR",
```
auphelia: Just as a remark, if we're changing the builder argument to be a node attribute, this would need to be changed here.

auphelia (changes requested):
Overall this looks good to me, let's have a chat about the builder argument vs. node attribute situation and then we can merge this into dev.

klassen9:
Hi @auphelia 

The latest commit contains the changes you requested.

For automatically assigning the FINNLoop's memory type, I added a DataflowBuildConfig resolver that runs within step_assign_ddr_weight_offsets, so no additional build step is needed.

Users can still override this automatic assignment by setting mem_type in the folding config, provided both options are available.

auphelia (approved):
Thank you @klassen9 ! Looks great and it's amazing that MLO is now also usable for embedded boards.

PR #1608, opened 2026-06-29
A parallel, self-balancing CI pipeline for FINN [CI 8/9]

merkelmarrow:
This is PR 8 of 9 of a series intended to make CI faster and more robust.

This PR overhauls FINN's main Jenkins build job with a parallel build pipeline. Instead of one long serial run, the test matrix is split into shards that run side by side across the available build agents, and the pipeline learns from past runs to keep those shards balanced over time.

The pipeline itself delegates as much logic as possible to Python. All of the real logic (such as which tests run, how they are sharded, how long they took last time, what to keep on disk, how to clean up) lives in the dependency-free `finn_ci` Python package that can be unit tested easily. The Jenkins side calls into it frequently. This is the main strategy shift.

## Why this matters

- **Much faster CI.** Work that used to run in sequence now runs in parallel, cutting CI times by about 75%.
-  **Self-balancing.** The pipeline records how long each group of tests takes, and uses that history to spread work evenly across shards automatically.
- **Reliability and observability.** The diff is very large. However, a very large proportion of this is mechanical error handling and error observability, which is the difference between a CI pipeline that can be repaired easily and a flaky one that will be abandoned. These robustness measures were developed and refined in response to real failure modes observed during over 80 Jenkins runs over the span of 3 months.
- **Testability.** CI logic now mainly lives in testable Python, and configuration drift between the pipeline definition, the docs, and the Python source of truth is caught by a unit test.

## Using it

For most contributors, nothing changes. You still write tests with the same markers, and you can run exactly what CI runs locally with a regular pytest invocation through Docker. There is a README guide for contributors and maintainers included with this change.

For Jenkins operators, a single shared-storage directory is the only additional setting needed. With it set, the pipeline gets the shared image cache, persistent timing history, per-agent download caches and hardware bitstream handoff. Retention is handled automatically, with old images, build artifacts and timing snapshots pruned according to the pruning policy in `finn_ci.retention`.

This change also adds some optional helpers for LSF integration.

## Behaviour

All logic has been tried and tested in a real lab environment. The failure paths learned from shared NFS, flaky build agents, orphaned LSF jobs, aborted builds etc. are much of the value of the pipeline. On failure, it surfaces the relevant tool logs and per-test failure summaries directly in the build output, and fails fast with clear errors when something can be caught at the validation stage. 

The pipeline is now much more resilient and flexible to unknown variables: it can safely delegate tool calls to a HPC cluster, or run everything locally, adapt to any number of build machines, use networked storage or fall back to local scratch.

It also lays the groundwork for the HW Jenkinsfile. Successful builds aggregate one package per board with a readiness marker that a later hardware pipeline will pick up (PR 9 of this series). 

## Breaking changes

As it stands, `Jenkinsfile_HW` has not been moved or changed. This is so that it can continue running with any saved artifacts in the legacy `ARTIFACT_DIR` until the migration is complete. The location of the CI, Brevitas, and full Jenkinsfiles will need to be updated in the relevant DSL jobs on the deployment side (they were moved to the `ci/` subdirectory). This is a breaking change for anyone running Jenkins with FINN. Otherwise, there are no other breaking changes, and only added functionality by configuring a shared storage location (`FINN_CI_NFS_ROOT`) on your deployment.

## Test/review

I recommend reading the included README during review. The `finn_ci` package is unit tested without a FINN install.  To run the tests locally:

```bash
python -m pytest tests/util/test_finn_ci_*.py tests/util/test_ci_config_sync.py
```


ci/finn_ci/retention.py
auphelia, line 1: The `prune_*` functions delete directories based on paths derived from FINN_CI_NFS_ROOT. Have you considered adding a sanity check to guard against accidental misconfiguration?
For example, if someone mistakenly sets FINN_CI_NFS_ROOT=/ or /home, the pruning could target unintended directories.

A simple check rejecting shallow paths or known system prefixes might help catch honest mistakes. Might be overkill since this is admin-configured, but curious what you think.

ci/Jenkinsfile_Brevitas
auphelia, line 1: Do you think we can adjust this Jenkinsfile to use the new setup as well? This Jenkinsfile has never been used but the intention with it was to be able to more often test with recent Brevitas changes, because we tend to stay quite long with older version (pinned by fetch-repos.sh) and it would be great to be more dynamic with upgrades.
Doesn't need to be part of this PR.

ci/finn_ci/retention.py
merkelmarrow: I think it's still quite safe since as you said, the operator would have to purposefully set it to that path, and secondly, the prune functions are very targeted to subdirectories that have the job key appended.

For instance, the image prune needs to walk FINN_CI_NFS_ROOT/docker_images/<job_key>. And inside that directory, the prunes are very targeted. For instance, image/artifact pruning only matches subdirs that are numerical digits only (inside of job_key) and snapshot prune only matches specific json files. And job_key is sanitised so it can't cause a tree traversal.

So in reality it would be perfectly fine for the operator to set the nfs root to their home directory.

auphelia: Thanks, sounds good!

ci/Jenkinsfile_Brevitas
merkelmarrow: I actually had in mind to delete both the CI and Brevitas Jenkinsfiles and making them parameters of the full pipeline instead as part of PR 9. What do you think? They can still be configured as separate jobs in DSL (they'll show separately on the Jenkins GUI)

run-docker.sh
```diff
@@ -259,6 +302,16 @@ DOCKER_EXEC+="-e LD_PRELOAD=/lib/x86_64-linux-gnu/libudev.so.1 "
 # Workaround for running multiple Vivado instances simultaneously, see:
 # https://adaptivesupport.amd.com/s/article/63253?language=en_US
 DOCKER_EXEC+="-e XILINX_LOCAL_USER_DATA=no "
+# Optional host cache for torch.hub / huggingface weights to avoid CDN 504s
+# on parallel CI runs. Bind target is /finn_cache (NOT $HOME, because docker
+# creates bind parents as root and that would break pip install --user).
+: ${FINN_DOCKER_CACHE_DIR=""}
+if [ -n "$FINN_DOCKER_CACHE_DIR" ]; then
+  mkdir -p "$FINN_DOCKER_CACHE_DIR/torch" "$FINN_DOCKER_CACHE_DIR/huggingface"
```
auphelia, line 310: Can you explain what this is actually for? We are not using torch.hub or huggingface in the FINN codebase. What's the use case for this cache? Is this based on actual CDN 504s you encountered, or is it for something for the future?

ci/Jenkinsfile_Brevitas
auphelia: I'll reach out offline about that, I think that Jenkinsfile_CI should stay independent. For the Brevitas one there might be a case for integration.

auphelia (approved):
This is amazing work @merkelmarrow ! Thanks a lot :) I will merge when you have reacted to my question about the huggingface cache.

run-docker.sh
merkelmarrow: Some of the tests pull example networks from Brevitas which in turn pull some torch.hub weights, I don't remember exactly which tests do (I think the mobilenet/end2end ones?) Let me get back to you on that. But the cache is active and working already. I _did_ see CDN 504s earlier in April-ish. I can't definitively say that it was due to pulling weights but it was a reasonable mitigation at the time to not pull weights on N shards on every run, so I kept it.

The huggingface cache is purely forward-facing at the moment, but I think it's reasonable. There's a good chance some test ends up pulling something from huggingface in the future, and for a couple of extra lines of code (while we're already setting up the torch cache), this will kick in automatically so someone doesn't have to figure the same thing out in a year's time when CI breaks with network errors. I had a look and some of the Brevitas examples already use huggingface: https://github.com/Xilinx/brevitas/tree/master/src/brevitas_examples/llm https://github.com/Xilinx/brevitas/tree/master/src/brevitas_examples/stable_diffusion

I'm happy to remove or clarify the comments if you would like.

auphelia: Thanks for the explanation! I think we can leave it like that for now.

auphelia (approved):
Thank you @merkelmarrow ! This looks good to me.

PR #1609, opened 2026-06-30
[Tests] Tolerate extra robust_rmtree retry from real NFS removal

merkelmarrow:
A robust_rmtree test is very rarely flaky because it stubs a failed tree removal due to ENOTEMPTY (NFS race) twice, then calls a real shutil.rmtree and asserts that "calls == 3". But that third call might fail with a real ENOTEMPTY. This patch changes the assertion to >= 3 calls to remove this transient issue.


auphelia:
Thank you @merkelmarrow !

PR #1611, opened 2026-07-06
dwc wide streams

ollycassidy13:
Allows flattened DWC folding for wide stream cases.
Groups the Python-side data-width-converter change separately from later HLS splitting work.
Change type: Python

### DIFF
--- a/src/finn/custom_op/fpgadataflow/hls/streamingdatawidthconverter_hls.py
+++ b/src/finn/custom_op/fpgadataflow/hls/streamingdatawidthconverter_hls.py
@@ -73,9 +73,7 @@ class StreamingDataWidthConverter_hls(StreamingDataWidthConverter, HLSBackend):
                 numMiddleWords % (outWidth // elemWidth) == 0
             ), "Error in DWC element-width split calculation"
             self.code_gen_dict["$DEFINES$"].append("#define IntermediateWidth %d" % elemWidth)
-            self.code_gen_dict["$DEFINES$"].append(
-                "#define NumMiddleWords %d" % numMiddleWords
-            )
+            self.code_gen_dict["$DEFINES$"].append("#define NumMiddleWords %d" % numMiddleWords)
         elif self.needs_lcm():
             lcmWidth = self.get_iowidth_lcm()
             assert numInWords % (lcmWidth / inWidth) == 0, "Error in DWC LCM calculation"
@@ -98,8 +96,7 @@ class StreamingDataWidthConverter_hls(StreamingDataWidthConverter, HLSBackend):
         if self._needs_element_width_split():
             self.code_gen_dict["$DOCOMPUTE$"] = [
                 'hls::stream<ap_uint<IntermediateWidth>> intermediate ("intermediate");',
-                "%s<InWidth, IntermediateWidth, NumInWords>(in0_V, intermediate, numReps);"
-                % op,
+                "%s<InWidth, IntermediateWidth, NumInWords>(in0_V, intermediate, numReps);" % op,
                 "%s<IntermediateWidth, OutWidth, NumMiddleWords>(intermediate, out0_V, numReps);"
                 % op,
             ]
--- a/src/finn/custom_op/fpgadataflow/hls/streamingdatawidthconverter_hls.py
+++ b/src/finn/custom_op/fpgadataflow/hls/streamingdatawidthconverter_hls.py
@@ -41,6 +41,9 @@ class StreamingDataWidthConverter_hls(StreamingDataWidthConverter, HLSBackend):
     """Class that corresponds to finn-hlslib StreamingDataWidthConverter_Batch
     function."""
 
+    def _needs_element_width_split(self):
+        return self.needs_lcm() and self.get_iowidth_lcm() > 8191
+
     def get_nodeattr_types(self):
         my_attrs = {}
         my_attrs.update(StreamingDataWidthConverter.get_nodeattr_types(self))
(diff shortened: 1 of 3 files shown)


ollycassidy13:
- Enables DWC folding when a stream word spans more than the tensor’s channel dimension.
  - Avoids oversized HLS intermediate integer types for non-divisible wide-stream conversions.
  - Keeps the logical tensor data and shape unchanged; only the folded stream representation changes.
  - The folded-shape fix is shared by HLS and RTL DWCs, while the oversized-LCM fallback is HLS-specific.

src/finn/custom_op/fpgadataflow/hls/streamingdatawidthconverter_hls.py
```diff
@@ -41,6 +41,9 @@ class StreamingDataWidthConverter_hls(StreamingDataWidthConverter, HLSBackend):
     """Class that corresponds to finn-hlslib StreamingDataWidthConverter_Batch
     function."""
 
+    def _needs_element_width_split(self):
+        return self.needs_lcm() and self.get_iowidth_lcm() > 8191
```
STFleming, line 45: It might be good to have a test parameter in `test_fpgadataflow_dwc.py` that is exploring this case to make sure it's behaviour is expected.

STFleming (changes requested):
Thanks @ollycassidy13 

All dwc tests that I tried were passing for me with this which is great. However, I don't believe any of them were testing the latest capabilities that this PR added. Would it be possible to update some of the parameters of `test_fpgadataflow_dwc.py` so that we are testing those parameters.

ollycassidy13:
# FINN DWC Wide-Stream Changes

This change makes FINN's `StreamingDataWidthConverter` (DWC) handle stream words that are wider than a tensor row, while also avoiding excessively wide intermediate types in generated HLS.

A DWC does not change the logical tensor. It repacks the same ordered sequence of bits into stream words of a different width.

## 1. Folded-shape fix

Previously, FINN assumed that every stream word fit inside the tensor's final dimension, which is usually the channel dimension. It calculated the folded shape as:

```python
[..., channels // elements_per_word, elements_per_word]
```

That assumption fails for the following case:

```text
shape       = [1, 2, 8]
datatype    = INT2
input width = 32 bits = 16 elements
```

The last tensor dimension contains only eight elements, but one 32-bit input word contains sixteen `INT2` elements. A single stream word therefore spans both eight-element rows. The old calculation effectively attempted to produce:

```text
[1, 2, 8 // 16, 16] = [1, 2, 0, 16]
```

This is not a valid representation of the tensor.

The new `_folded_shape_for_width()` method has two paths:

- If the number of elements per stream word divides the final tensor dimension, FINN preserves the outer tensor dimensions as before.
- Otherwise, FINN flattens the complete tensor and groups the flattened element sequence into stream words.

The example tensor contains sixteen elements in total, so its 32-bit input folded shape becomes:

```text
(1, 16)
```

This represents one stream word containing the entire 32-bit tensor. The logical tensor shape remains `[1, 2, 8]`; only its transport representation changes.

The fallback also checks that the stream width contains a whole number of datatype elements and that the total number of tensor elements can be divided into complete stream words. Partial words or implicit padding are not introduced.

## 2. Why the HLS converter normally uses an LCM

The underlying `StreamingDataWidthConverter_Batch` HLS template only directly supports conversions for which one width divides the other. Its compile-time condition is equivalent to:

```cpp
InWidth % OutWidth == 0 || OutWidth % InWidth == 0
```

For widths such as 128 and 130 bits, neither width divides the other. FINN normally handles this by converting through their least common multiple (LCM):

```text
128 bits -> 8,320 bits -> 130 bits
```

This is mathematically valid because:

```text
8,320 / 128 = 65
8,320 / 130 = 64
```

The first converter combines 65 input words into one LCM-width word. The second converter splits that word into 64 output words. The ordered bit sequence is unchanged.

However, this approach requires the HLS implementation to create an internal type and stream such as:

```cpp
ap_uint<8320>
hls::stream<ap_uint<8320>> intermediate;
```

The change treats an LCM width above 8,191 bits as oversized:

```python
self.needs_lcm() and self.get_iowidth_lcm() > 8191
```

Such conversions use the alternative path described below instead of generating the oversized LCM intermediate.

## 3. Oversized-LCM fallback

Instead of converting upward through the LCM, the fallback converts downward through the datatype's element width:

```text
input width -> element width -> output width
```

For the BINARY regression case, each element is one bit wide, so the conversion becomes:

```text
128 bits -> 1 bit -> 130 bits
```

The test tensor has shape `[1, 65, 128]` and contains:

```text
1 * 65 * 128 = 8,320 bits
```

The corresponding word counts are:

- `8,320 / 128 = 65` input words.
- `65 * (128 / 1) = 8,320` one-bit intermediate words.
- `8,320 / 130 = 64` output words.

The number of intermediate words is calculated as:

```python
numMiddleWords = numInWords * (inWidth // elemWidth)
```

The generated HLS is conceptually equivalent to:

```cpp
hls::stream<ap_uint<1>> intermediate("intermediate");

StreamingDataWidthConverter_Batch<128, 1, 65>(
    in0_V, intermediate, numReps
);

StreamingDataWidthConverter_Batch<1, 130, 8320>(
    intermediate, out0_V, numReps
);
```

This satisfies the HLS template's divisibility rule in both stages:

- The element width divides the input width.
- The element width divides the output width.

The added assertions verify that both stream widths are aligned to complete datatype elements and that the intermediate word count can form an exact number of output words.

The two converter stages communicate through an internal HLS stream and execute as a dataflow region. They preserve the original bit order and do not change the logical tensor contents or shape.

The fallback prioritizes a simple and guaranteed element-aligned conversion. Its trade-off is throughput: a very narrow intermediate stream, such as the one-bit stream in this example, can require many cycles. A wider common divisor could potentially improve performance further.

PR #1612, opened 2026-07-06
Limit XSI xelab threading

ollycassidy13:
Limits XSI `xelab` threading in the FINN XSI adapter.
This is a small simulation-build stability change for resource-heavy RTL simulations.
Change type: Python


STFleming (changes requested):
Thanks @ollycassidy13. This looks good to me. Only minor change I would add is we probably need to document the ENV VAR usage somewhere.

STFleming (approved):

auphelia:
Thanks for the PR, @ollycassidy13! 
The SIGABRT issue with xelab's auto-threading on shared servers is definitely worth fixing.

One thought: I'm a bit hesitant to add another env var here. We already have `NUM_DEFAULT_WORKERS` which other processes in FINN use for parallelism. Could we just use that directly instead of introducing `FINN_XELAB_MT`?
That way xelab respects the same setting as everything else, and users who already tune `NUM_DEFAULT_WORKERS` for their environment get consistent behavior across all FINN tools.

Or am I maybe overlooking something? Is there a benefit to keeping xelab threading independent from the general worker count that I'm not seeing?

ollycassidy13:
Thanks! `NUM_DEFAULT_WORKERS` is already used as the fallback when `FINN_XELAB_MT` is unset, so existing configurations get the consistent behaviour you describe.

The reason for retaining an optional xelab-specific override is that the two settings control different levels of parallelism: `NUM_DEFAULT_WORKERS` controls the number of FINN worker processes, while `xelab --mt` controls threads within each xelab process. For example, `PrepareRTLSim` can run several node-level elaborations in parallel, so using the same value for both can result in multiple multithreaded xelab processes and approximately quadratic oversubscription. xelab also tends to be more memory- and stability-sensitive than the other parallel transformations.

My intent was therefore for `NUM_DEFAULT_WORKERS` to remain the normal default, with `FINN_XELAB_MT` available only when users need to cap xelab more tightly without serializing the rest of FINN. Therefore I think it best to retain it as an optional override?

auphelia:
Thanks for the quick reply! That's a fair point, I hadn't considered that these control two different levels of parallelism. I will quickly check that the documentation formats correctly on the rtd website and then I think this PR is ready to be merged.

PR #1613, opened 2026-07-06
Allow padded XSI input feed for FIFO sim

ollycassidy13:
Allows padded XSI input feeds during FIFO simulation.
This is a low-level XSI runtime compatibility fix for stream padding behavior.
Change type: C++

### DIFF
--- a/finn_xsi/rtlsim_xsi.cpp
+++ b/finn_xsi/rtlsim_xsi.cpp
@@ -17,9 +17,19 @@
 #include <vector>
 #include <tuple>
 #include <functional>
+#include <cstdlib>
 
 #include "xsi_finn.hpp"
 #include "rtlsim_config.hpp"
+
+static unsigned env_unsigned(char const *name, unsigned const fallback) {
+	char const *const  value = std::getenv(name);
+	if(!value || !*value)  return fallback;
+
+	char *end = nullptr;
+	unsigned long const  parsed = std::strtoul(value, &end, 10);
+	return end == value? fallback : static_cast<unsigned>(parsed);
+}
 
 int main(int argc, char *argv[]) {
 
@@ -43,6 +53,8 @@ int main(int argc, char *argv[]) {
 		size_t  itodo = istream_descs.size();
 		size_t  otodo = ostream_descs.size();
 		size_t  omute = ostream_descs.size();
+		unsigned const  input_n_inferences =
+			n_inferences + env_unsigned("FINN_XSI_INPUT_PAD_INFERENCES", 0);
 
 		// Find I/O Streams and initialize their Status
 		struct stream_status {
@@ -126,6 +138,10 @@ int main(int argc, char *argv[]) {
 
 		// Start Stream Feed and Capture
 		std::cout << "Starting data feed with idle-output timeout of " << max_iters << " cycles ...\n" << std::endl;
+		if(input_n_inferences != n_inferences) {
+			std::cout << "Feeding " << input_n_inferences << " input frame(s) for "
+				<< n_inferences << " expected output frame(s)." << std::endl;
+		}
 
 		// Make all Inputs valid & all Outputs ready
 		for(auto &s : istreams)  s.port_vld.set(1).write_back();
@@ -146,15 +162,15 @@ int main(int argc, char *argv[]) {
 				bool const  rdy = s.port_rdy.read()[0];
 				if(vld && !rdy)  continue;
 
-				// Track successgul Transactions
+				// Track successful Transactions
 				if(vld) {
 					s.job_txns++;
-					if(++s.total_txns == s.job_size * n_inferences)  itodo--;
+					if(++s.total_txns == s.job_size * input_n_inferences)  itodo--;
 				}
 
 				// Proceed according to Throttling Rate
 				if((s.job_txns < s.job_size) || !(iters < s.await_iter)) {
-					if(s.total_txns < s.job_size * n_inferences) {
+					if(s.total_txns < s.job_size * input_n_inferences) {
 						if(!vld)  to_write.emplace_back(s.port_vld.set(1));
 						if(s.job_txns == s.job_size) {
 							s.job_txns = 0;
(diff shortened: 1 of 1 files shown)


STFleming (changes requested):
Thanks @ollycassidy13.

It looks like this is increasing the number of inferences that are pushed into `rtlsim_xsi.cpp` which is used during FIFO depth simulation. 

I would argue that this is already a feature built into FINN with the `fifosim_n_inferences` builder argument. https://github.com/Xilinx/finn/blob/68f789ba0eb1229c041be9ef900759fcb1650284/src/finn/builder/build_dataflow_config.py#L304

However maybe I'm overlooking a use case that requires this? 

It also looks like this is increasing the number of input inferences without increasing the checking logic for how many outputs are observed and they get flagged as `spurious outputs` and things will error out.

ollycassidy13:
The requirement is to feed `N + padding` inputs but capture exactly `N` outputs; setting `fifosim_n_inferences` to `N + padding` would also wait for the padded outputs and is not equivalent.

For a single output stream, the current code exits as soon as the Nth output completes, so it will not observe a padded output as spurious. This could be changed to handle multiple output streams if required (deassert each output stream’s `tready` as soon as it reaches `N`)?

STFleming:
Thanks that clears things up a bit @ollycassidy13!

I think I'm finding the term padding a bit misleading here, to me this would mean resizing within a tensor/frame of a single inference to add some padding, but this is essentially increasing the number of input inferences (i.e. it increases the number of complete input frames) if I understand it correctly? 

I'm also still a little unsure **why** you need to feed in `N + padding` inputs and collect only `N` outputs instead? Can you explain the motivation for the PR a bit more please.

ollycassidy13:
The FIFO characterization needs continued input traffic while the first inference traverses the pipeline so it can observe overlapping activity and peak FIFO occupancy. Using only one input frame removes input traffic too early. Setting `fifosim_n_inferences=2` is not equivalent because it also requires a second output, and the TinyDeiT loop body stalled in the tail of that second dummy inference.

The intended behavior is:
1. feed 2 complete input frames
2. measure FIFO activity during the overlap
3. finish after receiving 1 complete output frame

ollycassidy13:
# XSI Extra Input Frames for FIFO Characterization

## Why this is needed

FINN's FIFO-sizing simulation inserts large monitored FIFOs, runs the stitched design in XSI, and records each FIFO's peak occupancy through its `maxcount` signal. These peak occupancy measurements are then used to choose suitable FIFO depths.

For the measurement to be representative, the simulator needs to exercise the pipeline under overlapping traffic. With only one input frame, input traffic stops as soon as that frame has entered the design:

```text
input frame 1  ------------------|
                                  `- input becomes idle
pipeline              processing frame 1 ------------------>
```

As the first frame moves through the pipeline, upstream stages become idle and internal FIFOs begin draining. The observed `maxcount` values may therefore describe an isolated, draining inference rather than the occupancy that occurs under sustained traffic.

Feeding another complete input frame keeps the source active while the first frame is still moving through the pipeline:

```text
input       frame 1 ------ frame 2 ------------------------->
pipeline       frame 1 ------------------------------------>
                         frame 2 ---------------------------->
FIFO activity          <--- overlapping traffic --->
```

This overlap can produce additional backpressure and greater FIFO occupancy. Those higher occupancies are precisely what the FIFO-sizing pass needs to observe; otherwise, it can underestimate the FIFO depth required during continuous operation.

## Why `fifosim_n_inferences=2` is not equivalent

The existing `fifosim_n_inferences` setting couples the number of input frames with the number of output frames the simulation must receive:

```text
fifosim_n_inferences = 2

feed 2 input frames
wait for 2 output frames
```

For a long TinyDeiT/FINNLoop XSI simulation, waiting for both output frames can add substantial runtime. The second dummy inference supplies the desired overlapping input activity near the start of the run, but its output is not needed for FIFO sizing. Requiring that dummy inference to propagate through the complete design keeps the simulator running after the useful FIFO occupancy measurement has already taken place.

The required characterization window is instead:

```text
feed up to 2 complete input frames
measure FIFO activity during their overlap
finish after receiving 1 complete output frame
```

The additional frame is stimulus used to keep the pipeline loaded. It is not an additional output that the characterization run needs to collect.

Separating the two targets therefore shortens the characterization run without removing the overlapping traffic used to measure peak FIFO occupancy. This matters particularly for large stitched designs, where XSI simulation is slow and completing even one additional end-to-end inference can take a significant amount of wall-clock time.

## Intended behavior

The XSI input-feed extension separates the input traffic target from the output completion target:

```text
input target  = N + extra input frames
output target = N
```

For the TinyDeiT example:

```text
N                  = 1
extra input frames = 1

input target       = 2 frames
output target      = 1 frame
```

The simulator continues presenting input traffic for the additional frame while it waits for the requested output. Once the first output frame completes, the FIFO occupancy measurement is complete and the simulation may finish without waiting for the dummy frame to propagate all the way to the output.

This behavior allows FINN to observe the FIFO pressure caused by overlapping inferences and then stop once the requested measurement window is complete. It avoids spending simulation time carrying an otherwise unnecessary dummy inference all the way to the output.

auphelia:
Thanks for putting this together @ollycassidy13, and for the detailed rationale on sustaining traffic during FIFO measurement.

After discussing it with the team, we've decided to close this one. The main reasoning:

- As implemented, the padding is effectively a runtime optimization rather than a new measurement capability. The input-side traffic under FINN_XSI_INPUT_PAD_INFERENCES=k is equivalent to setting n_inferences = N+k; the FIFO occupancy being measured is the same. The difference is that this exits once N outputs are collected rather than draining the extra frames' outputs, which saves tail latency on deep designs, but overlaps with the existing n_inferences knob for the measurement itself.
- It's gated behind another manual env var and isn't wired into the FIFO characterization flow, so the benefit doesn't apply automatically.
- More importantly, we're expecting analytical FIFO sizing (#1185) to address the underlying FIFO-sim pain points this is trying to help with, which reduces the need for rtlsim-characterization tweaks like this going forward.

To be clear, closing this doesn't lose the work, the PR and its history remain fully searchable and visitable here, and we can revisit or pick it back up if a concrete use case comes up that n_inferences doesn't cover (e.g. measuring steady-state behaviour without paying for output propagation). Thanks again for the contribution and the write-up.

PR #1615, opened 2026-07-06
Add optional rtlsim stage debug logging

ollycassidy13:
Adds optional stage-level debug logging to `rtlsim_exec`.
The behavior is gated by `FINN_RTLSIM_DEBUG_STAGES` and is isolated from normal simulation output.
Change type: Python


STFleming (approved):
Thanks @ollycassidy13.

This PR adds some debug printing on the various stages of executing an rtlsim.

The only thing I'm unsure about is if this should using `logging` like in  `build_dataflow` or if it should be kept as a print where maybe it is nice to see the output printed in the terminal as it is running. I can see the arguments for both. 

I'm a little unsure of the ENV variable for controlling this, I'd be tempted just to have it print on default. If we did go for an ENV var it would need to be documented better somewhere.

auphelia:
This is useful for internal debugging, but I'm not sure it belongs in the main codebase. It adds another environment variable to an already large pool, and the scope is very narrow (just tracing rtlsim stages).

For users who need this level of debugging, adding print statements locally is straightforward. Would it make more sense to document this as a debugging technique rather than adding permanent instrumentation?

ollycassidy13:
Thanks for the feedback. I’ve removed `FINN_RTLSIM_DEBUG_STAGES` and moved the stage messages to the standard Python logger at DEBUG level, so there is no new environment variable or default terminal noise.

I think retaining the instrumentation is worthwhile because these specific failures occur frequently enough during long processes and the resulting error often does not identify where the failure happened. These lightweight stage records make CI and user reports much easier to diagnose without requiring local source edits each time.

auphelia:
Thanks for putting this together @ollycassidy13, and for iterating on the logging vs env-var approach based on the earlier feedback.

After digging into it, though, I don't think the change does what it's intended to as written. The stage messages are logged through finn.core.rtlsim_exec at DEBUG level, but nothing in FINN attaches a handler to that logger (or to finn/root), and the build_dataflow.log handler lives on the separate build_dataflow logger with propagate=False. So the records propagate up to root, find no handler, and fall through to Python's last-resort handler, which is fixed at WARNING. The DEBUG lines get silently dropped. In a normal build or a standalone rtlsim run they don't end up anywhere: not in build_dataflow.log, and not on the console. They'd only surface if someone manually configured a DEBUG handler on the root logger (e.g. logging.basicConfig(level=logging.DEBUG)), which nothing in the flow does at the moment.

The unit test passes because pytest's caplog fixture injects its own capturing handler and forces the level to DEBUG for the block, so it verifies the log call is made, but not that the messages reach anywhere a user or CI would actually see them. That means the main motivation here (easier CI/user-report diagnosis without local source edits) doesn't really hold in practice.

Beyond that, the coarse stage markers overlap with the newer debug FIFO infrastructure, which covers the "it hung and I don't know where" cases more thoroughly, and in practice none of us reached for this kind of setup even during our recent heavy FIFO-sim debugging.

For those reasons I'm going to close this. If there's a concrete case where the stage-level visibility is worth keeping, the way to actually deliver it would be to emit at INFO from within the rtlsim verification build step so the messages land in build_dataflow.log automatically, happy to reopen and go that route if we find it's needed. Thanks again for the contribution.

PR #1616, opened 2026-07-06
Compact RTL elementwise broadcast memstreams

ollycassidy13:
Compacts RTL elementwise broadcast parameter memstreams and replays them through FINNLoop tap repetition.
Includes the operator, FINNLoop integration and focused test update for the same behavior.
Change type: Python


src/finn/custom_op/fpgadataflow/rtl/elementwise_binary_rtl.py
```diff
@@ -482,20 +482,11 @@ def make_weight_file(self, weights, weight_file_mode, weight_file_name):
...
-            else:
-                tile_factor = int(num_w_reps)
-            weight_tensor = np.tile(
-                weight_tensor, (tile_factor,) + (1,) * (len(weight_tensor.shape) - 1)
-            )
+        if "decoupled" in weight_file_mode:
+            weight_tensor = weight_tensor.reshape(1, -1, weight_tensor.shape[-1]).copy()
+            if weight_file_mode == "decoupled_npy":
+                np.save(weight_file_name, weight_tensor)
+                return
```
STFleming: Just to flag for merging, this looks like a conflict/overlap with #1618. Either we merge these changes in this PR and close #1618 or we can merge #1618 before merging this. 

(#1618 LGTM btw)

STFleming (approved):
Great work @ollycassidy13 thanks for this!

I had found this issue also and was starting to address it in `bugfix/eltwise_rtl_mlo_memdepth` but your solution is a lot better I believe. 

I think the `TAP_REPS` approach also makes sense here, but I'd like to get @auphelia's thoughts on that. Also, I just wanted to flag that there is some overlap with the PR #1618 so we might need to take a bit of care when merging this.

auphelia (approved):
Thanks @ollycassidy13.
Approved with minor changes incorporated during review.

Changes incorporated during review

1. Simplified make_weight_file() in elementwise_binary_rtl.py
2. Removed redundant unit test test_elementwise_rtl_broadcast_const_memstream_length
3. Added RTL elementwise to FINNLoop MLO end-to-end test

Note: This PR supersedes #1618, which will be closed after merge.

PR #1617, opened 2026-07-06
Include nested FINNLoop sources for stitched rtlsim

ollycassidy13:
Adds missing nested FINNLoop HDL sources to the stitched-IP rtlsim source list.
Includes the source-list helper and its focused IP-stitch regression test.
Change type: Python

### DIFF
--- a/src/finn/transformation/fpgadataflow/create_stitched_ip.py
+++ b/src/finn/transformation/fpgadataflow/create_stitched_ip.py
@@ -40,8 +40,62 @@ from shutil import copytree
 from finn.transformation.fpgadataflow.replace_verilog_relpaths import (
     ReplaceVerilogRelPaths,
 )
-from finn.util.basic import make_build_dir, resolve_xilinx_tool
+from finn.util.basic import getHWCustomOp, make_build_dir, resolve_xilinx_tool
 from finn.util.fpgadataflow import is_hls_node, is_rtl_node
+
+
+RTLSIM_SOURCE_EXTENSIONS = {".v", ".sv", ".vh", ".svh", ".vhd"}
+
+
+def append_missing_finnloop_rtlsim_sources(model, v_file_list):
+    """Add nested FINNLoop-generated HDL sources to the top rtlsim source list.
+
+    Vivado's top-level USED_IN_SYNTHESIS query can omit HDL generated under a
+    packaged FINNLoop IP's own block design. The XSI compile then sees the
+    top-level wrapper but misses nested MLO helper modules such as generated
+    FIFOs, DuplicateStreams, and loop-body shuffles. The FINNLoop IP-generation
+    step already exports a complete local rtlsim source list, so merge entries
+    that are not already present in the top source list.
+    """
+
+    if not os.path.isfile(v_file_list):
+        return
+
+    with open(v_file_list) as f:
+        existing_sources = [line.strip() for line in f if line.strip()]
+    existing_paths = set(existing_sources)
+    existing_basenames = {os.path.basename(path) for path in existing_sources}
+    appended_sources = []
+
+    for node in model.graph.node:
+        if node.op_type != "FINNLoop":
+            continue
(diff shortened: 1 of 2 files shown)


src/finn/transformation/fpgadataflow/create_stitched_ip.py
```diff
@@ -40,9 +40,59 @@
 from finn.transformation.fpgadataflow.replace_verilog_relpaths import (
     ReplaceVerilogRelPaths,
 )
-from finn.util.basic import make_build_dir, resolve_xilinx_tool
+from finn.util.basic import getHWCustomOp, make_build_dir, resolve_xilinx_tool
```
STFleming, line 43: Same issue as in #1621 for `getHWCustomOp`

src/finn/transformation/fpgadataflow/create_stitched_ip.py
```diff
@@ -40,9 +40,59 @@
...
 from finn.util.fpgadataflow import is_hls_node, is_rtl_node
 
+RTLSIM_SOURCE_EXTENSIONS = {".v", ".sv", ".vh", ".svh", ".vhd"}
+
+
+def append_missing_finnloop_rtlsim_sources(model, v_file_list):
+    """Add nested FINNLoop-generated HDL sources to the top rtlsim source list.
+
+    Vivado's top-level USED_IN_SYNTHESIS query can omit HDL generated under a
```
STFleming, line 52: When you say "can omit" is this a random sporadic error that you were observing? Or is it consistently missing sources in a particular scenario? 

The reason I ask is because on `finn/dev` the `test_finnloop_end2end_mlo` tests are passing, wouldn't we expect these tests to fail if there were sources missing, perhaps this has already been addressed on `dev`?

tests/fpgadataflow/test_fpgadataflow_ipstitch.py
```diff
@@ -61,6 +65,105 @@
...
+
+
+class FakeModel:
+    def __init__(self, nodes):
+        self.graph = FakeGraph(nodes)
+
+
+@pytest.mark.fpgadataflow
+def test_ipstitch_appends_missing_finnloop_rtlsim_sources(tmp_path, monkeypatch):
+    top_list = tmp_path / "all_verilog_srcs.txt"
```
STFleming, line 95: This test proves that for a hand-constructed set of source lists the helper appends the expected HDL entries. 

However, what I think would be a stronger test is justifying that the helper function is needed. What would be stronger and prove this is an integration test where we build a real FINNLoop stitched IP and assert that after CreateStitchedIP the compiled XSI object contains / can elaborate the nested modules. Ideally this test would show that without the call to `append_missing_finnloop_rtlsim_sources` it is failing and then when the call is added it is passing.

tests/fpgadataflow/test_fpgadataflow_ipstitch.py
```diff
@@ -61,6 +65,105 @@
...
+
+    model = FakeModel(
+        [
+            FakeNode("MVAU_rtl", str(loop_a)),
+            FakeNode("FINNLoop", str(loop_a)),
+            FakeNode("FINNLoop", str(loop_b)),
+            FakeNode("FINNLoop", str(tmp_path / "missing_loop")),
+        ]
+    )
+    monkeypatch.setattr(
```
STFleming, line 149: This monkey patching might disappear after fixing `getHWCustomOp`

src/finn/transformation/fpgadataflow/create_stitched_ip.py
STFleming, line 52: When you say "can omit" is this something that happens unpredictably with the underlying tools, or it will always always happen in specific scenarios?

STFleming (changes requested):
Thanks for the PR @ollycassidy13!
I've left some comments inline, please reach out if anything doesn't make sense.

PR #1618, opened 2026-07-06
Fix RTL elementwise MLO parameter npy generation

ollycassidy13:
Fixes RTL elementwise MLO parameter NPY generation for loop-body parameter streams.
This is a narrow operator/runtime-data fix for elementwise binary RTL inside FINNLoop.
Change type: Python


STFleming (approved):
This looks great to me, thanks for the contribution @ollycassidy13!

I think if you run a standalone `execute_node` you don't see this issue as it overwrites the numpy file correctly, but if you don't run that you hit this issue. Thanks for the fix.

auphelia:
Closed, see https://github.com/Xilinx/finn/pull/1616#pullrequestreview-4752704660

PR #1619, opened 2026-07-06
RTL softmax updates

ollycassidy13:
Carries the RTL softmax HDL/template and Python wrapper updates from finnlib.
Change type: Python and RTL


STFleming:
Thanks @ollycassidy13! 

It looks like this was a partial update to the Softmax RTL variant HWCustomOp which was not up on dev yet. 
I've got the RTL Softmax up on `feature/softmax-rtl` so can you change this PR to now target #1624 instead of dev? Thanks!

PR #1620, opened 2026-07-06
Pad1d operator

ollycassidy13:
Add generic pad1d operator to FINN.
This also replaces AddCLSToken as the functionality is absorbed into this more general module.
Change type: Python and RTL


src/finn/custom_op/fpgadataflow/rtl/pad1d_rtl.py
```diff
@@ -0,0 +1,206 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
STFleming: Use new lic format please

src/finn/custom_op/fpgadataflow/rtl/pad1d_rtl.py
```diff
@@ -0,0 +1,206 @@
...
+            "PAD_RIGHT_DATA": self._pack_pad_data(model, 2),
+        }
+
+        for key, value in code_gen_dict.items():
+            template = template.replace("$%s$" % key, str(value))
+
+        code_gen_dir = self.get_nodeattr("code_gen_dir_ipgen")
+        with open(os.path.join(code_gen_dir, topname + ".v"), "w") as f:
+            f.write(template)
+        shutil.copy(rtlsrc + "/pad1d.sv", code_gen_dir)
```
STFleming: Copying here is slightly different to how other rtl operations do things (such as [thresholding_rtl](https://github.com/Xilinx/finn/blob/924e2e72d66bda90e7f97073a73044d5b99f9f28/src/finn/custom_op/fpgadataflow/rtl/thresholding_rtl.py#L431)) which do it as part of the tcl operation, is either way okay @auphelia or is it best to stick to the way it's done previously?

src/finn/custom_op/fpgadataflow/pad1d.py
```diff
@@ -0,0 +1,250 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
STFleming: Use new lic style please

finn-rtllib/pad1d/hdl/pad1d_template.v
```diff
@@ -0,0 +1,84 @@
+/******************************************************************************
```
STFleming: New lic header style please

src/finn/custom_op/fpgadataflow/rtlbackend.py
```diff
@@ -85,14 +85,21 @@ def code_generation_ipi(self):
     def code_generation_ipgen(self, model, fpgapart, clk):
         self.generate_hdl(model, fpgapart, clk)
 
+    def get_rtlsim_input_indices(self):
```
STFleming: Modifying the rtlbackend.py for a custom op addition worries me a little. I can see why you did it, but generally when you have non-standard input handing the `execute_node` would be overridden to handle it. This is arguably a useful generalisation but it feels like if it was to be added it should be in it's own PR. I'd like to get @auphelia's opinion on this one.

src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -1334,6 +1334,164 @@ def apply(self, model):
...
+            if idt is None or not idt.is_integer():
+                continue
+
+            const_dtypes_valid = True
+            for inp in node.input:
+                if inp == stream_name:
+                    continue
+                pad_dt = model.get_tensor_datatype(inp)
+                if pad_dt is None:
+                    model.set_tensor_datatype(inp, idt)
```
STFleming: I believe this `set_tensor_datatype` needs to be deferred until later. We could have a situation where we perform a modification but then bail later (line 1452) and move onto the next node. I think the solution here is to add this to a list and modify once we are sure this is the node we want infer on.

tests/fpgadataflow/test_fpgadataflow_pad1d.py
```diff
@@ -0,0 +1,437 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
```
STFleming: new lic header

tests/fpgadataflow/test_fpgadataflow_pad1d.py
STFleming, line 1: Can we try applying the test standardising skill to this please.

STFleming (changes requested):
Thanks for this @ollycassidy13, on the whole this is a great PR for a very useful operator. I have some minor concerns that I've highlighted, some of them I'd like to get @auphelia's opinion on, such as the modifications to the `rtlbackend.py`

STFleming:
It also looks like the current tests are failing due to an import error 
```
________ ERROR collecting tests/fpgadataflow/test_fpgadataflow_pad1d.py ________
ImportError while importing test module '/home/runner/work/finn/finn/tests/fpgadataflow/test_fpgadataflow_pad1d.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/fpgadataflow/test_fpgadataflow_pad1d.py:53: in <module>
    from finn.transformation.fpgadataflow.synth_ooc import SynthOutOfContext
E   ModuleNotFoundError: No module named 'finn.transformation.fpgadataflow.synth_ooc'
```

src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -1334,6 +1334,167 @@ def apply(self, model):
...
 
 
+class InferPad1DLayer(Transformation):
+    """Convert 1D Concat padding patterns into Pad1D.
+
+    This covers class-token insertion as ``Concat([cls_token, tokens], axis=1)``
+    as well as constant left/right 1D padding around one streamed tensor.
+    """
+
+    def _get_tensor_shape(self, model, tensor_name):
```
auphelia: The `_get_tensor_shape` helper seems unnecessary. The transformation already classifies inputs via `model.get_initializer(inp) is not None`, so at each call site we know whether we're dealing with a tensor or an initializer. Just use the appropriate method directly.
Additionally, the fallback could mask unexpected input patterns. If inputs don't match what the transformation expects, we'd want an explicit failure rather than a silent workaround.
Suggest removing the helper and using direct lookups matching the expected type at each call site:

```
# For streamed input (known to be dynamic tensor)
stream_shape = model.get_tensor_shape(stream_input)

# For constant input (already verified as initializer)
init = model.get_initializer(const_input)
const_shape = list(init.shape)
```

src/finn/custom_op/fpgadataflow/pad1d.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def get_instream_width(self, ind=0):
+        if ind != 0:
+            return 0
+        return self.get_input_datatype().bitwidth() * self.get_nodeattr("SIMD")
+
+    def get_outstream_width(self, ind=0):
+        return self.get_output_datatype().bitwidth() * self.get_nodeattr("SIMD")
+
+    def get_number_output_values(self):
```
auphelia: `get_number_output_values `can be removed. It's identical to the parent implementation in HWCustomOp (the int() cast should be unnecessary).

tests/fpgadataflow/test_fpgadataflow_pad1d.py
auphelia, line 1: Thanks for the initial cleanup on the test file. I think we can consolidate further to reduce duplication and test the full pipeline more naturally.

Have `test_fpgadataflow_pad1d` and `test_pad1d_stitched_ip_synth_ooc` start from a Concat model rather than a pre-built Pad1D node. This way they test the complete flow: Concat → InferPad1DLayer → SpecializeLayers → rtlsim. Resource estimation checks can also be folded into the main test (like you have already in the ooc synthesis test).

This would let us remove:

  - `test_pad1d_rtl_codegen`: If rtlsim passes with correct output, codegen is implicitly verified. Checking specific hex strings in generated Verilog is fragile and redundant.
  - `test_pad1d_python_execution_with_repeated_padding`: The repeated padding case (pad_left=2 with single-token input) is already covered by the (2, 2, 1) config in the main test.
  - `test_convert_concat_does_not_mutate_rejected_node`: This tests that failed pattern matching doesn't corrupt the graph, but we don't have similar tests for other ops. If apply() returns modified=False, the graph
   shouldn't change. This is expected behavior, not something specific to Pad1D.
  - `test_convert_concat_to_pad1d`: Covered by the main tests starting from Concat.

This would leave us with 2 focused tests covering the full flow rather than 7 with overlapping coverage.

auphelia (changes requested):
Thanks for the Pad1D operator implementation @ollycassidy13. The overall structure looks great. Few comments for cleanup, see inline comments for details.

auphelia:
> It also looks like the current tests are failing due to an import error
> 
> ```
> ________ ERROR collecting tests/fpgadataflow/test_fpgadataflow_pad1d.py ________
> ImportError while importing test module '/home/runner/work/finn/finn/tests/fpgadataflow/test_fpgadataflow_pad1d.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> /usr/lib/python3.10/importlib/__init__.py:126: in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
> tests/fpgadataflow/test_fpgadataflow_pad1d.py:53: in <module>
>     from finn.transformation.fpgadataflow.synth_ooc import SynthOutOfContext
> E   ModuleNotFoundError: No module named 'finn.transformation.fpgadataflow.synth_ooc'
> ```

This was addressed by merging `upstream/dev`.

auphelia (approved):
Thank you @ollycassidy13 ! Looks great :)

STFleming (approved):
Looking good, thanks @ollycassidy13!

PR #1622, opened 2026-07-07
[FINN LOOP] Adding a guard to the thresholding so that the adj_list i…

STFleming:
This PR adds an extra guard statement for thresholding layers when building the adjacency list.


auphelia (approved):
Thanks @STFleming !

PR #1623, opened 2026-07-08
Sim Engine timing issue

STFleming:
Changings to the sim engine timings were causing issues with the DSP58 tests. The initial values returned were corrupted. Restoring it to the original timings restores passing tests.


auphelia (approved):
Thank you, @STFleming ! Looks good to me, we will do additional testing on the dev branch but thanks a lot for bringing this hotfix in.

PR #1624, opened 2026-07-08
RTL Softmax variant

STFleming:
Adds an RTL softmax primitive `finn-rtllib/softmax_rtl` plus a `softmax_rtl.py` `HWCustomOp` and expands the tests to include the RTL variant. The hardware architecture is a streaming softmax over fixed-length FP32 vectors of N elements consumed at SIMD elements/beat (N/SIMD beats per vector). Builds on the `pwpolyf` module for the exponential calculation and uses an `int_to_fp32.sv` primitive to allow for flexible input data types.

Currently this is restricted to versal architectures only as it relies on the DSP58 hardware primitive. 

**Pipeline stages** (free-running, elastic)
  1. `softmaxf_max` — SIMD reduction tree + time-domain max accum + replay buffer; emits (max, has_infty).
  2. `softmaxf_exp` — per-lane FP SUB(x-max), polynomial exp (via pwpolyf), ufixed conversion, SIMD sum tree + time-domain sum accum.
  3. `softmaxf_recip` — Newton–Raphson reciprocal of sum (FP32→FP32, configurable iters).
  4. `softmaxf_div` — per-lane ufixed→FP32, FP MUL by reciprocal.
  
This overall seems to get better resource consumption and better timing closure than the current HLS implementation, with an estimated fmax of ~400MHz for SIMD < 8.  

image


src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -1807,8 +1807,6 @@ def apply(self, model):
                     name=n.name,
                     SIMD=1,
                     NumChannels=input_shape[-1],
-                    cpp_interface="hls_vector",
-                    hls_style="freerunning",
```
STFleming, line 1811: Do these need to be added in during specialise layers for the HLS variant?

auphelia (approved):
Thanks @STFleming, looks good to me! I actually removed the dsp combination conflict test file, I think we've added this to check the different cases but it's actually quite difficult to integrate them into the test suite, because some of them fail with segfaults which corrupts succeeding tests. I will see if I can come up with a cleaner testing approach in a separate PR.

PR #1625, opened 2026-07-09
Improve data packing speed

klassen9:
Rewrites `finnpy_to_packed_bytearray` to pack with vectorized NumPy operations instead of the per-element hex-string path, and removes the now-redundant `fast_mode` flag. Fast paths are selected automatically from the input dtype/container.

The new packer is benchmarked against the legacy one on a folded input shape `(1, 3, 224, 56, 4)` (N, H, W, FOLD, SIMD), i.e. a 224×224×3 image folded with `simd=4`. Each dtype is run through both versions across both reverse flags.

The new packer is orders of magnitude faster across most dtypes (seconds to sub-millisecond), with the biggest wins on whole-byte dtypes where packing reduces to a byte view. However, the increase in generality compared to the legacy `fast_mode` leads to a decline in packing speed for the INT8/UINT8 case. (The speedup comparison of the float32 containers to the `fast_mode` cases is unfair, as they always assume a native container; it is only included for completeness. Nevertheless, the old `fast_mode` still slightly outperforms the native-container case of the new implementation.)

packing_speedup

## Note: output may be non-contiguous

The fast paths pack by returning views and flips of the input rather than copying. That's what makes them fast, but it means the returned array can be non-contiguous, and for the byte-view path it shares memory with the input.

This is safe for the intended consumers: the driver stages packed data into a pre-allocated contiguous DMA buffer (`allocate` + `np.copyto`) before sending it to the device, so the view is materialized on copy. Any caller that needs a standalone, contiguous buffer, e.g. feeding the result straight back into `packed_bytearray_to_finnpy`, or writing it to a file, must wrap it in `np.ascontiguousarray` (or others such as `np.copy`) first. The test suite was updated to do exactly this on the unpack side.


auphelia (approved):
Thank you @klassen9 ! Looks good to me.

PR #1630, opened 2026-07-15
Update the SLASH_ALVEO flow for SLASH's renamed slashkit linker

merkelmarrow:
The V80 SLASH_ALVEO backend invokes the SLASH linker as "v80++". The current SLASH release (v1.0) renamed that tool to "slashkit", so SlashLink fails to link. This PR updates every instance of v80++ to slashkit and wires it through the heavy tool resolver (`resolve_xilinx_tool`).

This is a breaking change for:

- anyone using pre-v1.0 SLASH with FINN's SlashLink (I don't think that's anyone at the moment).
- anyone running a CI setup that shells out heavy processes to a compute farm (`FINN_TOOL_DIR_OVERRIDE`). A new slashkit shim needs to be added to the tool dir overrides, all slashkit invocations by FINN will fail in such a CI setup until this is done.

Two related dockerfile issues were also fixed:

- The linker deb is now installed after an `apt-get update`, standard best practice and slashkit declares apt dependencies such as python3-jinja2 that the image doesn't already install, so the local deb install needs current package lists (would fail against stale lists cached in an earlier layer).
- The final `USER` references the literal `USERNAME` instead of `$USERNAME` so the image default user was an account that doesn't exist (this affects any image that is layered *from* FINN's). This was fixed.

## Testing

- [x] New unit tests in `tests/fpgadataflow/test_slash_link.py` that test command construction for hw/sim, kernel list assembly.
- [x] One xfailing test added because SlashLink doesn't yet pass the synthesis clock period to the linker (follow-up work).
- [x] E2E test ongoing.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -263,6 +263,6 @@ A site that offloads the heavy Xilinx tools to a compute farm (see "Running tool
 
 | Env var                  | What it sets                                                                                                                                        |
 | ------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `FINN_TOOL_DIR_OVERRIDE` | Shim directory. `finn.util.basic.resolve_xilinx_tool()` resolves `vivado`/`v++`/`vitis_hls`/`vitis-run`/`xelab` to `<dir>/<tool>` when set.        |
+| `FINN_TOOL_DIR_OVERRIDE` | Shim directory. `finn.util.basic.resolve_xilinx_tool()` resolves `vivado`/`v++`/`vitis_hls`/`vitis-run`/`xelab`/`slashkit` to `<dir>/<tool>` when set.        |
 
 The wrapper's own variables are deployment-specific.
--- a/docker/Dockerfile.finn
+++ b/docker/Dockerfile.finn
@@ -88,7 +88,7 @@ RUN locale-gen "en_US.UTF-8"
 # Install v80++. If the package isn't provided, this is a no-op
 # The wildcard pattern ensures COPY doesn't fail if the file doesn't exist
 COPY v80pp.de[b] /tmp/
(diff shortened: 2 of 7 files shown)


merkelmarrow:
E2E test was successful (SLASH_ALVEO build of the cybersecurity MLP to a `.vbin`).

merkelmarrow:
Want to improve the tests.

merkelmarrow:
Force push:

- rebased on dev
- improved slash tests and moved them to util (fpgadataflow test is coming)
- removed xfailing test since passing the parameter wouldn't do anything anyway (https://github.com/Xilinx/SLASH/issues/98)
- updated getting_started.rst
- added _slash_link_argv() wrapper so integration with the tool resolver is testable (`alveo_build.py`)

quetric:
>removed xfailing test since passing the parameter wouldn't do anything anyway (https://github.com/Xilinx/SLASH/issues/98)

@merkelmarrow This is a bug we're actively working to fix - see https://github.com/Xilinx/SLASH/pull/169 - perhaps just disable the test until the fix is merged.

merkelmarrow:
> > removed xfailing test since passing the parameter wouldn't do anything anyway ([Xilinx/SLASH#98](https://github.com/Xilinx/SLASH/issues/98))
> 
> @merkelmarrow This is a bug we're actively working to fix - see [Xilinx/SLASH#169](https://github.com/Xilinx/SLASH/pull/169) - perhaps just disable the test until the fix is merged.

The xfailing test was just an awkward reminder to add the feature later, thanks for fixing. I'll fold in the clock-hz arg for SlashLink in FINN later

auphelia (approved):
Thank you @merkelmarrow !

PR #1632, opened 2026-07-17
Fix steady-state RTLSIM throughput reporting

ollycassidy13:
# RTLSIM Throughput Reporting

## Problem

The RTLSIM performance report did not reliably distinguish end-to-end latency from steady-state throughput.

Latency and throughput require different measurements. If an accelerator produces output frames at completion times

```text
t1, t2, ..., tN
```

then `t1` measures the pipeline-fill latency of the first frame. Steady-state throughput must instead be derived from the spacing between completed output frames. A simulation that completes only one output frame has no output-to-output interval and therefore cannot measure steady-state throughput.

The previous reporting path had several related problems:

1. The default RTLSIM batch size was one frame. Despite having no measurable output interval, the report copied the one-frame batch throughput into `stable_throughput[images/s]`. This presented a latency-dominated result as though it were steady-state throughput.

2. For a batch of `N > 1` frames, the report removed the first-frame latency from the elapsed time but still used `N` as the throughput numerator. There are only `N - 1` output intervals between the first and last of `N` completed frames, so this overestimated throughput. The old calculation was effectively:

   ```text
   N / post-fill elapsed time
   ```

   whereas the required calculation is:

   ```text
   (N - 1) / (last output completion - first output completion)
   ```

3. The XSI simulation harness emitted `interval_cycles` without reporting whether an interval was valid or how many complete output frames had been observed. Downstream code therefore had no reliable way to reject an interval from a one-frame or incomplete run.

4. Input throttling state and output timing state shared storage through a union in the XSI harness. This made the timing fields overlap and prevented the harness from independently tracking the first completion, latest completion, latest interval and number of completed output frames.

5. The reporting code did not require the simulation to finish successfully before publishing a steady-state result. A timed-out run or a run with unfinished input or output streams could consequently yield a misleading throughput value.

The overall effect was that `stable_throughput[images/s]` could describe pipeline-fill behavior, use an incorrect frame count or be based on an incomplete measurement. It was therefore not a dependable statement of the accelerator's sustained frame rate.

### Explicit output-frame timing in the XSI harness

The shared input/output statistics union was removed. Each stream now has separate fields for input throttling and output measurement. For every output stream, the harness independently tracks:

- the cycle at which the first complete output frame is received;
- the cycle at which the most recent complete output frame is received;
- the most recent completion-to-completion interval; and
- the number of complete output frames received.

The first completed output records latency only. An interval is recorded only when a subsequent output frame completes.

The raw XSI result now includes:

- `completed_output_frames`: the number of complete frames observed across all output streams, limited by the least-complete output stream;
- `interval_valid`: whether at least two complete output frames were observed and a non-zero interval exists;
- `steady_state_frames`: the number of measured intervals, equal to `completed_output_frames - 1`; and
- `steady_state_cycles`: the first-to-last completion span used for sustained-throughput calculation.

For designs with multiple output streams, the reported completion count uses the minimum completed-frame count across the streams, while the timing span uses the maximum first-to-last span. This prevents a faster output stream from overstating the throughput of the complete multi-output result.

### Separation of batch and steady-state throughput

The report retains `throughput[images/s]` as the throughput of the complete simulated batch, including pipeline fill:

```text
batch_throughput = N * 1e9 / (clock_period_ns * total_cycles)
```

This value remains useful for latency-sensitive workloads, but it is not labelled as steady-state throughput.

Sustained throughput is now computed from the completed-frame span:

```text
steady_state_throughput = steady_state_frames * 1e9
                          / (clock_period_ns * steady_state_cycles)
```

For a complete batch of `N` frames, this is:

```text
steady_state_throughput = (N - 1) * 1e9
                          / (clock_period_ns * (tN - t1))
```

The report also exposes `fps_from_interval`, calculated from the latest valid output interval:

```text
fps_from_interval = 1e9 / (clock_period_ns * interval_cycles)
```

`fps_from_interval` describes one measured completion interval, while `stable_throughput[images/s]` averages all measured intervals between the first and last completed outputs.


tests/util/test_rtlsim_performance.py
auphelia, line 1: I really like the additional assertions in `test_fifosizing.py`. 
Checking `completed_output_frames`, `interval_valid`, `steady_state_frames/cycles`, and using an exact `pytest.approx` for `stable_throughput` is a solid improvement and would catch the kind of numerator regression this PR fixes.

I'm less convinced by most of `test_rtlsim_performance.py`. Tests such as `uses_completed_frame_span`, `corrects_legacy_multi_frame_numerator`, and the throughput assertion in `defaults` effectively inject `steady_state_frames`/`steady_state_cycles` and then verify the formula. They'd still pass if the underlying XSI accounting bug (`first_complete`/`last_complete` aliasing) were reintroduced, since they never exercise that path. I'd keep `rejects_single_frame_interval` and `rejects_incomplete_run`, as those validate actual behavior rather than arithmetic.

For the regression itself, could we instead extend `test_end2end_build_dataflow_directory` in `tests/util/test_build_dataflow.py`? It already runs the stitched-IP + rtlsim-performance flow. Running it with `rtlsim_batch_size >= 2` and asserting on `latency_cycles`, `steady_state_cycles` (first-to-last span), and `completed_output_frames` would directly exercise the multi-frame path that broke and fail if the harness accounting regressed.

Minor note: with the default `rtlsim_batch_size` bump to 2, does the example `dataflow_build_config.json` still explicitly set it to 1? The assertion on `report/rtlsim_perf_batch_1.wdb` looks like it assumes batch 1, so it's worth double-checking that remains valid.

auphelia (changes requested):

auphelia (approved):
Thank you @ollycassidy13 ! Looks good to me.

PR #1633, opened 2026-07-20
[Sim Engine] Adjust the cycle print rate and report on the liveness of each output

STFleming:
This PR is a small adjustment to the output printing while a `finnxsi` simulation is running. 
* Adjusts the cycle reporting from every cycle to every 10K cycles
* prints the current watchdog status for each output every 10K cycles so that it is easier to inspect how far away a timeout condition might be occurring.


auphelia (approved):
Looks good to me, thanks @STFleming !

PR #1637, opened 2026-07-22
Feature/expanded finn examples

lstasytis:
This PR is an extension of the https://github.com/Xilinx/finn/tree/feature/add_finnexamples branch with the primary goal of becoming a testing harness for the folding, DWC insertion and fifo sizing flows in FINN to make it easier to perform end-to-end tests.

The idea is to provide some ground truths in json format that the flows have to match to test that they work. We can offer two separate modes for each test: slow and fast. Fast would only run up to estimate-report step, avoiding all vivado calls and just testing all the finn machinery for these flows. The slow flow would additionally perform synthesis for all the 3 flows and rtlsim for the fifo sizing one so that we can get a complete test.

Will iterate on this in the coming days.


tests/benchmark/_feature_bench.py
```diff
@@ -0,0 +1,258 @@
...
+    if fifo_nodes:
+        for n in fifo_nodes:
+            inst = getCustomOp(n)
+            bits += inst.get_nodeattr("depth") * inst.get_instream_width()
+    else:
+        for n in model.graph.node:
+            inst = getCustomOp(n)
+            try:
+                depths = inst.get_nodeattr("outFIFODepths")
+            except (AttributeError, Exception):  # noqa: B014 - attr may be absent
```
auphelia, line 210: `except (AttributeError, Exception`) is equivalent to `except Exception`. Exception already subsumes AttributeError, so the # noqa: B014 is muting something that should be flagged. 

But more importantly, this silently skips any node whose FIFO storage exists but can't be read. Suggest catching only the specific expected case:
```
 try:
    depths = inst.get_nodeattr("outFIFODepths")
 except AttributeError:
    continue
```
and letting anything unexpected surface.

tests/benchmark/_e2e_bench.py
```diff
@@ -0,0 +1,417 @@
...
+model_id = fb.model_id
+
+
+def results_dir():
+    d = os.environ.get("FINN_E2E_RESULTS", os.path.join(REPO_ROOT, "e2e_results"))
+    os.makedirs(d, exist_ok=True)
+    return d
+
+
+def build_root():
```
auphelia, line 99: Two concerns here:
1. Build artifacts default into the repo tree (REPO_ROOT/e2e_build, and REPO_ROOT/e2e_results in results_dir()). FINN's test guideline is to use scratch/make_build_dir() locations, not the working tree. At minimum these dirs need to be in .gitignore so a stray git add can't commit build output. The /tmp-is-full rationale in the docstring is environment-specific and shouldn't drive mainline defaults.
2. os.environ.setdefault("FINN_BUILD_DIR", ...) mutates the process environment as an import/collection-time side effect of a helper. Prefer setting this via monkeypatch.setenv inside a fixture so it's scoped and parallel-safe.

tests/benchmark/_e2e_bench.py
```diff
@@ -0,0 +1,417 @@
...
+
+
+# ---- running + metric collection --------------------------------------------
+
+
+def run_e2e_build(model_file, cfg):
+    """Cached rtlsim build. Returns wall-clock seconds (0.0 for a cache hit)."""
+    import finn.builder.build_dataflow as build
+
+    os.chdir(REPO_ROOT)
```
auphelia, line 301: `os.chdir()` changes the working directory for the whole process, which is not safe under pytest-xdist and can leak into unrelated tests. Since the builds already use absolute paths (cfg.output_dir, model_file), consider dropping the chdir entirely, or if a specific CWD is genuinely required, scope it with monkeypatch.chdir() in a fixture.

tests/benchmark/e2e_report.py
```diff
@@ -0,0 +1,158 @@
...
+    for path in sorted(glob.glob(os.path.join(results_dir, "*__*.json"))):
+        with open(path) as f:
+            r = json.load(f)
+        if "model" not in r or "flow" not in r:
+            continue  # experiment dumps and other non-flow JSONs
+        by_model.setdefault(r["model"], {})[r["flow"]] = r
+    return by_model
+
+
+def _tp(r):
```
auphelia, line 49: This is a near-verbatim duplicate of stable_throughput() in _e2e_bench.py:372 (same fclk/interval_cycles logic, same fallback chain, same "builder's stable_throughput degenerates" rationale in the comment). Two copies of the steady-state-throughput definition will drift. If the metric is ever refined, one side gets missed and the report silently disagrees with the regression assertion. Suggest exporting the single implementation from _e2e_bench (or wherever the shared scaffolding lands) and importing it here.

auphelia (changes requested):
Thanks @lstasytis  for putting this together. 

Before diving into specifics, my main concern is that this PR is very bloated for what the base branch sets out to do (add finn-examples build tests). I'd like us to find a leaner shape, and I'll follow up with concrete suggestions on that.

A couple of cross-cutting style points that recur throughout:

- Module naming isn't FINN style. The leading-underscore support modules (_feature_bench.py, _e2e_bench.py) aren't a convention used elsewhere in the tree. An underscore usually marks a package-private member, but these are standalone modules loaded via sys.path.insert(...). Moving the shared scaffolding into a conftest.py (or plainly named modules) under tests/benchmark/ would fix the naming and let you drop the repeated sys.path.insert(0, ...) + # noqa: E402 boilerplate at the top of every test file. You could also use util/test.py
- Copyright headers are inconsistent. Please use the two-liner we have in new files (without year and the full text): 
```
```
I've left inline comments on specific lines, and more suggestions will follow once we've aligned on the overall scope.

PR #1638, opened 2026-07-23
Trim PYNQ driver dependencies

auphelia:
The generated PYNQ driver copies a handful of qonnx/finn source files into the  deployment folder. Two of them were copied verbatim and pulled heavy dependencies onto the board that the driver never uses:
- `qonnx.util.basic` imports `onnx` (only needed for `gen_finn_dt_tensor` + `roundup_to_integer_multiple`)
- `finn.util.data_packing` imports `bitstring` (only needed for `finnpy_to_packed_bytearray` / `packed_bytearray_to_finnpy` and their helpers, none of which touch `bitstring`)

This meant `onnx` and `bitstring` had to be installed on the target, which is a hassle for no functional benefit.

## Changes
- `make_driver.py` now emits trimmed-down versions of these two modules via `inspect.getsource`, containing only the functions the driver actually needs.
- Added a fast smoke test that verifies the generated modules import and round-trip with `onnx`/`bitstring` made unavailable, and contain no such imports.

### DIFF
--- a/src/finn/transformation/fpgadataflow/make_driver.py
+++ b/src/finn/transformation/fpgadataflow/make_driver.py
@@ -27,10 +27,12 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import inspect
 import json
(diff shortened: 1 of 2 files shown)


auphelia:
@klassen9, could you please have a look? I unfortunately can't mark you as a reviewer on GitHub.

klassen9:
Looks good to me! I verified it by generating a deploy package and running it on an actual board.

PR #1639, opened 2026-07-24
Crop RTL and SelectToken Operator

ollycassidy13:
Adds RTL implementation for Crop and the SelectToken fpgadataflow layer, converting supported Gather patterns into a hardware op for selecting tokens from a token sequence.


auphelia (approved):
Really nice work, @ollycassidy13! Thanks for this. 

I made one small change on top: I replaced the get_nodeattr override in SelectToken_rtl with an overridable _get_template_param_dict() hook on Crop_rtl. The override was synthesizing Crop attributes on lookup, which created a get/set asymmetry for attributes that don't physically exist on the node. Now SelectToken_rtl just overrides that hook to map its token geometry onto the crop core, same RTL reuse, but normal attribute semantics stay intact. I also dropped the hardcoded lut_estimation=200 (inherits the base like Crop now).

Thanks again!

PR #1641, opened 2026-07-24
[Tests] Isolate shared scratch and process-global test state

merkelmarrow:
The existing test suite leaks files and environment mutations into globally accessible state, which make different tests interfere with each other when run concurrently under xdist.

This is the mentioned follow-up to https://github.com/Xilinx/finn/pull/1599 . This PR systematically eliminates the remaining unsafe practices and adds a new test guidelines README that will enforce these going forward.

### List of changes

**Scratch paths**

- move runtime-weight serialisation, configuration exports, and temporary JSON files into unique build directories
- introduce a shared runtime-weight-stream helper that cleans successful temporary output and retains it on failure
- remove the global deploy dir for the BNN tests and use normal build dir instead (deploy dir no longer necessary after https://github.com/Xilinx/finn/pull/1608 with find_copy_zip.sh).
- `test_fpgadataflow_ipstitch.py` no longer saves into the module-level `ip_stitch_model_dir`.

**Loop extraction template path**

`LoopExtraction` previously always wrote `loop-body-template.onnx` into the cwd which is racy with concurrent tests. It now has an optional `loop_body_template_path`:

- The dataflow builder passes `cfg.output_dir/loop-body-template.onnx` directly and no longer moves the file afterward.
- Without passing the arg, the caller gets a unique directory under `FINN_BUILD_DIR`.
- Loop-rolling tests pass an explicit path under their per-test build dir.

**Environment**

- Replace direct environment mutation with monkeypatch.

**Documentation**

- Switch `docker/quicktest.sh` and `docs/finn/developers.rst` from `--dist=loadfile` to `--dist=loadgroup`.
- Added `tests/README.md` with test guidelines going forward.

### Testing

Successful validation in Jenkins in finn_ci_testing job 94.

### Other

I renamed `FinnCiShardingPlugin` to `FinnCiPlugin` because I planned to add worker isolation logic but it ended up being too complicated for the benefit. But I still prefer the rename and it has no external effects.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -40,7 +40,7 @@ Decorate it with the existing markers. For example, `@pytest.mark.fpgadataflow`.
 
 ### Clean up scratch in a test
 
-Use FINN's `make_build_dir()` and tear it down with the `robust_rmtree()` helper.
+Do not write scratch into the current working directory. Use FINN's `make_build_dir()` and tear it down with the `robust_rmtree()` helper. Successful tests should remove disposable scratch, while failed tests can keep it for diagnosis.
 
 ### Add a new BNN parameter value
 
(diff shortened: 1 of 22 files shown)


auphelia (approved):
Thanks @merkelmarrow!

PR #1645, opened 2026-07-27
[Tests] Add per-item deterministic RNG seeding for pytest

merkelmarrow:
This change adds deterministic automatic seeding across all tests. Seeding is derived from each pytest node id which is individual to each test. This is done through an autouse fixture in `tests/conftest.py`. 

```python
# Example node id:
tests/fpgadataflow/test_foo.py::test_func_name[param_id]
```

This addresses problems we've historically had with determinism. This is a deliberate tradeoff; on the one hand we're losing the feedback that "one particular test fails 1% of the time, thus the reference output calculations aren't perfectly representative". But these kinds of failures add a lot of noise when we're more interested in finding actual regressions.

The intended RNG policy is described in the extension to the testing guidelines doc.

### Implementation

- 32 bit seed derived from `sha256(nodeid)` in `tests/rng_seed.py`. xdist loadgroup suffixes are stripped using the predefined regex in finn_ci.
-  `tests/conftest.py` adds an autouse fixture.
- some tests in `tests/util/test_rng_determinism.py` cover suffix normalisation and numpy / torch seed behaviour
- removed pinned seeds everywhere
- refactored a test that relies on weights used across multiple loops to compute the weights upfront (instead of resetting the seed and computing the same weights on each loop iteration: `test_fpgadataflow_convinputgenerator_rtl_dynamic.py`)

### Testing

- [x] a version of this change excluding minor changes was successful in Jenkins finn_ci_testing run 94.
- [x] individual tests files collect and run without problems in docker
- [x] new Jenkins run with latest changes

### Other notes

Maintainers, note the guideline in the README that `np.random.default_rng()` should be seeded with finn_test_seed. 

Tests that now fail due to tolerances at the deterministic seed should now be treated as deficient. Either the generator that produces the reference output should be improved to better model simulation outputs or the tolerances should be relaxed.


auphelia (approved):
Approved. Thanks for this @merkelmarrow !

A couple of small changes made on top:

- Moved the seed helper into the CI package: relocated rng_seed.py from tests/ into ci/finn_ci/ so it lives alongside the rest of the finn_ci support modules (it's already loaded via pytest_plugins/sys.path). Updated the two importers (tests/conftest.py and tests/util/test_rng_determinism.py) to `from finn_ci.rng_seed import ...`
- Copyright headers: trimmed the two newly added files (ci/finn_ci/rng_seed.py, tests/util/test_rng_determinism.py) to the current short form (no year / "All rights reserved").

No functional changes beyond the move. LGTM.

PR #1649, opened 2026-07-29
Separate InnerShuffle physical and logical input shapes

klassen9:
## Problem

When a Reshape was fused into a Shuffle, InnerShuffle ignored it and transposed the physical tensors last two axes. So (1,768,14,14) -> reshape -> (1,768,196) -> transpose (0,2,1) -> (1,196,768) swapped the two 14 dims instead of swapping 768 and 196, producing the wrong result.

## Fix

Track the logical (post-reshape) shape the transpose acts on separately from the physical input tensor
- in_shape physical tensor
- transpose_in_shape (new) logical view

Update: Also fixed the InnerShuffle page guard


STFleming (approved):
Great catch @klassen9 and thanks for bringing the fix up along with a test for the failing case!

I'm also happy for the page-guard fix to be included here also.

PR #1657, opened 2026-08-19
Requant Enhancements for Decoupled Parameters and Custom Signedness of Outputs

preusser-amd:
Extends Requant_rtl with two enhancements:

1. Signed output support: Requant_rtl now supports both signed and unsigned output clipping based on the output datatype. Previously only unsigned outputs were supported. 
2. Decoupled parameter mode: New mem_mode="internal_decoupled" option streams scale/bias/tap parameters from a unified memstream instead of constant-folding them into the RTL. This can be used standalone for runtime-updateable parameters, and is also a prerequisite for MLO support.

MLO Support

Building on decoupled mode, Requant can now be used inside FINNLoop bodies where parameters vary per loop iteration:
  - adapt_for_loop_body() auto-switches to decoupled mode when LoopRolling flags the node
  - Parameters are served by an indexed memstream with set-select driven by FINNLoop's stream tap graph

### DIFF
--- a/finn-rtllib/requant/hdl/requant.sv
+++ b/finn-rtllib/requant/hdl/requant.sv
@@ -1,6 +1,8 @@
 /****************************************************************************
- * Copyright Advanced Micro Devices, Inc.
- * SPDX-License-Identifier: MIT
+ * Copyright (C) 2026, Advanced Micro Devices, Inc.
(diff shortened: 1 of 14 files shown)


auphelia (approved):
Thanks for the contribution!

PR #1658, opened 2026-08-19
Sgerber.consolidated fifo

sgerber-amd:
Replaces different Fifo implementations with Thomas Preußers consolidated fifo component
which covers all memory styles. This implementation is centralized and reference only for
end users.

- The vivado codegen arm is removed, we rely solely on the new rtl component.
- `Q_srl` and `queue` implementations are transferred to fifo.
- `fifo.sv` has an `auto` mode for deciding implementation memory type based on depth.
   Alternatively, one can set the preferred memory type by setting RAM_STYLE for depths above 33, below it defaults to
   shift.
- added fitted resource estimator for all fifo implementations. This returns actual fifo
  resource + surrounding control resources (LUTs):
  The estimators are fit against 315 out-of-context syntheses (Vivado 2026.1; 134 on
  xczu7ev + 181 on xcvc1902, every style, both geometries, widths 1–256, depths 2–262145).
  Memory is exact on held-out sets;  LUT summed-absolute error is 4.3% (UltraScale+) / 5% (Versal)
- control signal fan out `.xdc` to be added as an optimization.
- `acc_stage.sv` actually changes implementation style. Q_srl was not really used as 
  a fifo, but as a rotator with the handshake tied off, now an SRL delay line.
- Some syntax changes that might break existing flows: 
            -  `impl_style` is now only `{"rtl"}`; other directives are ignored. 
            -  `SplitLargeFIFOs` / `get_adjusted_depth` /  `max_qsrl_depth` / `rtlsim_use_vivado_comps` removed; 
-if an architecture that does not support uram, vivado will ignore the uram directive and default to another fifo implementation.
- There is optimization potential in the fifo.sv decision tree between bram vs uram.


preusser-amd (approved):
Thanks so much, @sgerber-amd!
This is approved as much as the RTL is concerned.

auphelia (approved):
Thanks so much for this work @sgerber-amd and @preusser! 
Consolidating every FINN FIFO onto a single fifo.sv and propagating the change through the whole compiler is a significant refactor. 
You did a great job, I really appreciate how carefully you, @sgerber-amd, kept it non-disruptive.

I ended up being a bit more disruptive on top of your changes, since we're already touching all of this, it felt like the right moment to clean up some of the surrounding cruft rather than carry it forward. Main things I changed:

Retired now-dead attributes and config knobs
  - Dropped the `ram_style_resolved` and `is_versal` node attributes. `is_versal` is now derived from the `fpgapart` threaded into the estimators, matching how `dsp_estimation` already receives the part. I think this is a bit more future-proof if our resource modelling improves it will most likely be part dependent.
- Retired `impl_style` on StreamingFIFO_rtl (meaningless now that fifo.sv is the only backend). ApplyConfig tolerates a stale impl_style on a FIFO with a DeprecationWarning instead of erroring, so existing folding configs still load.
- Removed builder config `large_fifo_mem_style` as well 

Made the mirror harder to break
- Moved the FIFO cost/style model out of streamingfifo.py into a new finn/util/resource_models.py (a general home we can extend for other ops later) and pruned the comments down to the essential rationale.
- Added tests/fpgadataflow/test_fifo_ram_style_mirror.py, which parses fifo.sv's RAM_STYLE_EFF selection and fails if the Python _resolve() mirror drifts from the RTL. So, future edits to the selection can't silently desync the estimates.
- Added a warning when an explicit `ram_style` is requested at depth ≤ 33, where fifo.sv forces a shift register regardless.

Nice catch worth calling out
- Your fix in set_fifo_depths.py renaming the throwaway depth to tensor_size is a real bug fix. The output-FIFO sizing loop was previously reusing the input loop's stale value. Good spot.

None of this changes the RTL. Approving, thanks again for the effort and the care you put into keeping it clean.

PR #1660, opened 2026-08-20
Consolidating data width conversions.

preusser-amd:
Generalized RTL implementation of the data width converter.
Choosing this implementation more eagerly unless the user requests HLS explicitly.
Incorporating fix by @ollycassidy13 to mitigate wide intermediate `hls::stream` instantiations in the HLS implementation.


auphelia (approved):
Thank you, looks good to me!

PR #1662, opened 2026-08-24
[Build] Check subprocess return codes in synth/HLS launches

### LINKED ISSUE #52: Check status after script/command execution
Many transformations and utility functions in FINN call external commands such as `g++`, `vivado` and `vivado_hls` without checking the return status. If the called command fails, the FINN transformation won't be aware of this. Assertions should be introduced on the return status whenever an external command is called.

mumallaeng:
Fix #52.

## Problem

Several build steps launch Vivado, Vitis, vitis_hls, or g++ through `subprocess.Popen(bash_command, stdout=subprocess.PIPE); process.communicate()` and never check the return code. If the external tool fails, FINN doesn't notice: the build either continues with stale or missing artifacts, or fails later with a confusing error that isn't obviously related to the actual failure upstream.

`launch_process_helper` in `finn/util/basic.py` already exists to solve exactly this (it raises `CalledProcessError` when `check=True` and the process exits non-zero), but most of these call sites predate it and never got switched over.

## Fix

Route every remaining unchecked call site through `launch_process_helper(..., check=True)`:

- `finn/util/basic.py` - CppBuilder.build() (g++ compile)
- `finn/util/hls.py` - CallHLS.build() (vitis_hls)
- `finn/custom_op/fpgadataflow/hlsbackend.py` - exec_precompiled_singlenode_model()
- `finn/custom_op/fpgadataflow/rtl/finn_loop.py` - loop IP project build
- `finn/transformation/fpgadataflow/create_stitched_ip.py` - stitched IP project build
- `finn/transformation/fpgadataflow/make_zynq_proj.py` - Zynq synthesis
- `finn/transformation/fpgadataflow/alveo_build.py` - .xo packaging, Vitis link, and report-XML generation (3 call sites)
- `finn/core/rtlsim_exec.py` - 2 call sites already using launch_process_helper but not passing check=True

No behavior change on the success path. On failure, these now raise CalledProcessError at the point the tool actually failed, instead of silently proceeding.

**Update**: per review feedback, the call sites that had a follow-up context message (e.g. `"IPGen failed: %s not found"`) now wrap the `launch_process_helper` call in try/except, re-raising as `RuntimeError` with that message so the context isn't lost behind a bare `CalledProcessError`.
`launch_process_helper`'s `check` parameter also now defaults to `True`, so future call sites don't have to remember to pass it.

## Testing

isort/black/ruff (pinned versions from .pre-commit-config.yaml) pass on all changed files, and python -m py_compile passes. I don't have a licensed Vivado/Vitis toolchain in this environment to exercise these paths against a real tool failure; the change is a mechanical swap onto the same helper already used elsewhere in the codebase, e.g. alveo_build.py's SLASH link path already does the equivalent with `subprocess.run(..., check=True)`.

### DIFF
--- a/src/finn/core/rtlsim_exec.py
+++ b/src/finn/core/rtlsim_exec.py
@@ -295,7 +295,7 @@ def rtlsim_exec_cppxsi(
     # write compilation command to a file for easy re-running/debugging
     with open(sim_base + "/compile_rtlsim.sh", "w") as f:
         f.write(" ".join(build_cmd))
-    launch_process_helper(build_cmd, cwd=sim_base)
+    launch_process_helper(build_cmd, cwd=sim_base, check=True)
     assert os.path.isfile(sim_base + "/rtlsim_xsi"), "Failed to compile rtlsim executable"
 
     # launch the rtlsim executable
@@ -310,7 +310,7 @@ def rtlsim_exec_cppxsi(
             " ./rtlsim_xsi > rtlsim_xsi_log.txt"
             " 2> rtlsim_xsi_stderr.log"
         )
-    launch_process_helper(runsim_cmd, cwd=sim_base)
+    launch_process_helper(runsim_cmd, cwd=sim_base, check=True)
 
     # parse results file and return dict
     results_filename = sim_base + "/results.txt"
--- a/src/finn/custom_op/fpgadataflow/hlsbackend.py
+++ b/src/finn/custom_op/fpgadataflow/hlsbackend.py
@@ -30,14 +30,13 @@ import glob
 import numpy as np
 import os
 import re
-import subprocess
 import warnings
 from abc import ABC, abstractmethod
 from qonnx.core.datatype import DataType
 
 from finn import xsi
 from finn.custom_op.fpgadataflow import templates
-from finn.util.basic import CppBuilder, make_build_dir
+from finn.util.basic import CppBuilder, launch_process_helper, make_build_dir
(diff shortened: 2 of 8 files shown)


auphelia:
Hi @mumallaeng, thanks for this PR and thanks for your patience!

The issue you addressed is already quite old and over time we've added assert statements after subprocess calls to give users helpful context about what failed (e.g., "IPGen failed", "Vitis .xo file not created"). 

I still think your changes look good and address a real problem, because the error from the subprocess itself should be propagated. But one thing to consider: With check=True, the CalledProcessError is raised before reaching the mentioned asserts, so users would lose that context.

So, I would have two change suggestions for your PR:

1. Preserve helpful error messages if they are there: Wrap the calls in try/except to keep the context

```
try:
      launch_process_helper(bash_command, check=True)
except subprocess.CalledProcessError as e:
      raise RuntimeError(f"IPGen failed: {wrapper_filename} not found") from e
```

 2. Default to `check=True`: Consider making `check=True` the default in `launch_process_helper()` so future callsites don't accidentally forget it.

Thanks again for the contribution!

mumallaeng:
@auphelia Thanks for the review! 

Both suggestions are in now:

1. Wrapped the call sites that had a follow-up context message (IPGen failed, Vitis .xo/.xclbin not created, synthesis failed, etc.) in try/except, re-raising as RuntimeError with that message before the bare CalledProcessError reaches the caller.
2. Made `check=True` the default on `launch_process_helper`.

Ready for another look whenever you have time.

PR #1665, opened 2026-08-24
[Builder] Check that folding is configured for HW build outputs

### LINKED ISSUE #1430: Check that either folding config or target fps is set
## Prerequisites
None

### New behavior
There should be an error message early in the build flow if neither a folding config nor the target fps are specified in the build config. Therefore, checks could be added to ensure that either of them is always specified in the building config.

### Motivation
When starting the build flow with neither a folding config nor a target fps set, a weird error at the hw ipgen step occurs saying that the name of the projects hasn't been set.

### Parts of FINN affected
Transformation, FPGA Dataflow, Set Folding

mumallaeng:
Fix #1430.

## Problem

Building any hardware output (STITCHED_IP, BITFILE, OOC_SYNTH, RTLSIM_PERFORMANCE, PYNQ_DRIVER, DEPLOYMENT_PACKAGE) with neither `target_fps` nor `folding_config_file` set leaves every node at its creation-time `PE=1`/`SIMD=1` (`convert_to_hw_layers` sets this at node creation; `SetFolding` is what normally raises it toward a target). Nothing fails: the build just completes with the most folded, slowest possible implementation, and the misconfiguration only becomes obvious once the resulting accelerator turns out to be far slower than expected - after a full synthesis run.

I traced this through `SpecializeLayers` and `PrepareIP` with a minimal `PE=1`/`SIMD=1` MVAU model to confirm neither step raises on its own; the build genuinely proceeds silently rather than crashing.

## Fix

Add a `folding_missing` check to `run_all_config_checks` in `build_dataflow_checks.py`, following the same pattern already used for `zynq_board` and `vitis_platform` (pre-empting a known footgun with a clear message at config-check time, before the build reaches `step_hw_ipgen`). The check only fires when `generate_outputs` requests something beyond `ESTIMATE_REPORTS`, since estimate-only builds don't need folding to be explicitly configured.

Also updated `test_board_shell_compatibility`'s two success cases in `test_build_dataflow_checks.py`, which build `BITFILE` without `target_fps` - these were previously only exercising the board/shell check and would now (correctly) trip the new check too.

## Testing

Ran the full `test_build_dataflow_checks.py` suite locally (`setup-local.sh`-style venv, no Docker/Vivado) - all 15 tests pass, including 3 new ones covering the error case, the estimate-only pass-through case, and the target_fps pass-through case. `isort`/`black`/`ruff` (pinned versions from `.pre-commit-config.yaml`) pass on both changed files.

### DIFF
--- a/src/finn/builder/build_dataflow_checks.py
+++ b/src/finn/builder/build_dataflow_checks.py
@@ -266,6 +266,25 @@ def run_all_config_checks(cfg: DataflowBuildConfig) -> Report:
                 )
             )
 
+    needs_folding = cfg.generate_outputs and any(
+        o != DataflowOutputType.ESTIMATE_REPORTS for o in cfg.generate_outputs
+    )
+    if needs_folding and cfg.target_fps is None and cfg.folding_config_file is None:
+        checks.append(
+            _check(
+                "folding_missing",
+                Severity.ERROR,
+                False,
+                "Neither target_fps nor folding_config_file is set. Nodes stay at "
+                "their creation-time PE=1/SIMD=1, the most folded and slowest "
+                "implementation, and the build will silently run to completion "
+                "with it instead of failing fast",
+                "Set target_fps for automatic folding, or provide folding_config_file "
+                "for manual PE/SIMD settings, so a slow build isn't mistaken for "
+                "a deliberate choice",
+            )
+        )
+
     # === Environment Variables ===
     if has_bitfile and cfg.shell_flow_type == ShellFlowType.VITIS_ALVEO:
         missing = [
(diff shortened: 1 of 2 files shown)


auphelia:
Hi @mumallaeng,

Thanks a lot for your contributions and for opening the two PRs! We really appreciate the help.

We're currently in a bit of a crunch preparing for a project deadline and an upcoming conference, so reviews may take a little longer than usual. Please bear with us, we'll get to your PRs as soon as we can.

If you're planning to attend https://2026.fpl.org/, it'd be great to connect there as well!

Thanks again for your patience and contributions.

mumallaeng:
Hi @auphelia,

No worries at all, take your time - good luck with the deadline and the conference! I'm just glad these were useful; happy to wait however long the review takes.

I'd love to keep contributing and open more PRs if that's okay - if anything's urgent, just point me to it and I'll start there.

Unfortunately my schedule doesn't work out for FPL 2026 this time - maybe next time!

auphelia (approved):
Thanks @mumallaeng, looks good to me! This is a sensible guardrail that prevents users from accidentally building with default PE=1/SIMD=1 folding.

The logic correctly exempts estimate-only flows (where you might want quick estimates without folding optimization).

PR #1667, opened 2026-08-25
Improve LayerNorm reciprocal-square-root accuracy

ollycassidy13:
# Improve LayerNorm reciprocal-square-root accuracy

## Why

LayerNorm can operate on low-variance inputs where error in the existing reciprocal-square-root approximation becomes large enough to noticeably affect output accuracy in SigLip.

## How

This adds an optional second Newton–Raphson refinement:
`r = y × (1.5 - 0.5 × x × y²)`

The refinement is implemented as a pipelined three-FMA FP32 datapath. The LayerNorm configuration exposes `numRsqrtRefinements`, defaulting to the existing single-refinement behavior, and updates latency and DSP estimates accordingly. An RTL simulation test verifies the improved result against the ONNX LayerNorm reference.

### DIFF
--- a/finn-rtllib/layernorm/layernorm.sv
+++ b/finn-rtllib/layernorm/layernorm.sv
@@ -33,6 +33,7 @@
 module layernorm #(
 	int unsigned  N,
 	int unsigned  SIMD,
+	int unsigned  NUM_RSQRT_REFINEMENTS = 1,
 	bit  FORCE_BEHAVIORAL = 0
 )(
 	// Global Control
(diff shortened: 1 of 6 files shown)


auphelia:
Thanks for looking into the rsqrt accuracy issue. Currently in the demo, we’re handling the RTL/reference mismatch via a flag that makes the reference path match RTL behavior, so we can compare them consistently. We plan to assess the actual application-level accuracy impact separately before trading additional resources for accuracy improvements.

PR #1668, opened 2026-08-25
Fix threshold layout inference after transposes

ollycassidy13:
# Fix threshold layout inference after transposes

## Why

`InferThresholdingLayer` could interpret the input and output layouts of a `MultiThreshold` independently. After a non-canonical transpose, this could insert a one-sided layout conversion and silently reorder output values.

## How

Use one effective layout for both sides of the elementwise threshold operation. Shared thresholds follow the input layout, while per-channel thresholds follow the operator's declared layout.

### DIFF
--- a/src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
+++ b/src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
@@ -190,10 +190,24 @@ class InferThresholdingLayer(Transformation):
                 if not (tdt_int or tdt_fp or tdt_fxp):
                     continue
 
-                # check layout of inputs/outputs, and convert if needed
-                # check layout and convert if necessary
+                # Use one effective layout on both sides of MultiThreshold. The
+                # operation is elementwise and cannot change layout, but its
+                # default output annotation is NCHW even when a preceding
+                # transpose leaves the input layout non-canonical. Treating the
+                # two annotations independently would insert only an output
+                # transpose and silently reorder values. Per-tensor thresholds
+                # are channel-invariant, so they can use the existing layout;
+                # per-channel thresholds must follow the operator's declared
+                # data_layout.
                 thl_in_layout = model.get_tensor_layout(thl_input)
-                if thl_in_layout == DataLayout.NCHW:
+                shared_thresholds = thl_thres_shape[0] == 1
+                if shared_thresholds:
+                    convert_nchw = thl_in_layout == DataLayout.NCHW
+                else:
+                    multithreshold_layout = getCustomOp(node).get_nodeattr("data_layout")
+                    convert_nchw = multithreshold_layout == "NCHW"
+
+                if convert_nchw:
                     thl_input = nchw_to_nhwc(thl_input, model, node_ind)
                     node_ind += 1
                     thl_in_shape = model.get_tensor_shape(thl_input)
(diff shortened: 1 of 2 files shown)


auphelia:
Thanks for the PR, @ollycassidy13! 
 
I have some questions about the scenario that triggered this fix.

Which model/flow produced this pattern? Could you share what model or scenario led you to discover this bug? I'm trying to understand how a MultiThreshold node ends up with mismatched input/output layout annotations in practice.
My concern is that this fix might be masking an upstream bug.

Looking at this more closely, I think the root cause may be in InferDataLayouts (in qonnx), not InferThresholdingLayer. Here's my reasoning:

1. MultiThreshold is an elementwise operation, it doesn't change the data layout
2. Therefore, input and output layout annotations should always match
3. If they don't match, that's a bug in how InferDataLayouts handles MultiThreshold

Looking at qonnx/transformation/infer_data_layouts.py, I see that for MultiThreshold specifically, it reads the data_layout attribute to determine the output layout annotation, rather than propagating the input layout like it does for other custom ops. This seems incorrect for an elementwise op.

Suggestion

Rather than silently handling the mismatch in InferThresholdingLayer, would it make sense to:

1. Add an assertion that flags when input/output layouts don't match:
```
  if thl_in_layout != thl_out_layout:
      raise ValueError(
          f"MultiThreshold {node.name} has mismatched layouts: "
          f"input={thl_in_layout}, output={thl_out_layout}. "
          f"This indicates a bug in layout inference."
      )
```
2. Fix the root cause in qonnx's InferDataLayouts to propagate input layout for MultiThreshold (or update its data_layout attribute to match the actual input)

That said, if you have a real-world model that hits this issue and needs a workaround now, your fix does prevent incorrect graph generation. I'd just like to understand the scenario better before we decide on the right approach.

Tagging @iksnagreb, because I would like to hear his thoughts on this.

fpjentzsch:
You may want to look at how we currently do this on our end: https://github.com/eki-project/finn-plus/blob/c49e40ff9cfec7f88a6414e1432ba9b91fbb6a9d/src/finn/transformation/fpgadataflow/convert_to_hw/thresholding.py#L83

src/finn/transformation/fpgadataflow/convert_to_hw_layers.py
```diff
@@ -190,19 +190,32 @@ def apply(self, model):
...
+                # default output annotation is NCHW even when a preceding
+                # transpose leaves the input layout non-canonical. Treating the
+                # two annotations independently would insert only an output
+                # transpose and silently reorder values. Per-tensor thresholds
+                # are channel-invariant, so they can use the existing layout;
+                # per-channel thresholds must follow the operator's declared
+                # data_layout.
                 thl_in_layout = model.get_tensor_layout(thl_input)
-                if thl_in_layout == DataLayout.NCHW:
+                shared_thresholds = thl_thres_shape[0] == 1
```
iksnagreb, line 203: This correctly distinguishes per-tensor vs. per-channel thresholds for now, but does not generalize well to support arbitrary/multiple axes thresholds in the future (that's something I am working on for a while now). To make this more future-proof you could rather test for the product/volume of all leading dimensions being 1, i.e., something like `np.prod(thl_thres_shape[:-1]) == 1`.

iksnagreb:
Hm, at a glance I don't see anything inherently wrong with the proposed fix, but I am also concerned that this merely treats the symptoms of some bug (probably with the data layout propagation, which is known to be fragile and at least partially incorrect/guessing).

Eventually we should try to avoid further graph modifications to convert between layouts at this point in the flow - in this case the conversion was already there, you are just handling another edge-case, but overall we should stop adding more and more ad-hoc patches late into the flow and rather fail with a clear error message and start looking for the root cause... Here, we should then have a separate check for (a) input/output layout annotations to match (if present, if not, propagating input to output should be safe as this is kind of an elementwise operation, if they don't match that's a bug elsewhere) and (b) check whether the threshold parameter shape and `"data_layout"`  node attribute (if present) are consistent with these input/output layouts (actually I am not sure what to do if not and how/if it is even possible to end up in this situation, but that is also probably some bug elsewhere, likely in layout propagation as well).

PR #1669, opened 2026-08-25
Add enable to inner shuffle job tracker

ollycassidy13:
# Add enable to inner shuffle job tracker

Gate job completion tracking with `wr_en` to prevent false completions during stalled writes.


auphelia:
This looks sound to me, could you please provide a test case that fails on dev and passes with your changes applied?

preusser-amd (approved):
Thanks for tracking down and fixing this bug, @ollycassidy13!

auphelia:
Thank you @ollycassidy13 !

PR #1670, opened 2026-08-25
Propagate double-pumped FINNLoop clock

ollycassidy13:
# Propagate optional double-pumped clocks through FINNLoop

## Why

When a FINNLoop body contains double-pumped operators, its `clk2x` interface was not exposed or connected by the enclosing loop and stitched design.

## How

Detect the optional loop-body `clk2x` interface, propagate it through the FINNLoop wrapper and connect it during block-design stitching.


auphelia:
Looks reasonable overall. My main concern is that it enables double-pumped configurations that we don’t currently test (dynamic MVAU, fetch-weights + MVAU, etc.). Are you actually using double pumping in the SigLip body, and if so what verification coverage do we have for those paths?

PR #1671, opened 2026-08-25
Support oversized and single-buffered local weight memories

ollycassidy13:
# Support oversized and single-buffered local weight memories

## Why

Large external-weight buffers can exceed Vivado's per-array elaboration limit, while double buffering can consume more on-chip memory than necessary.

## How

Automatically split oversized weight memories into safe RAM banks, propagate the configured RAM style, and add an optional single-buffer implementation. Double buffering remains the default.


auphelia:
We’d like to better understand the actual problem being solved here. What limitation or failure were you seeing with the current implementation? 
Tagging @preusser and @d-kor

PR #1672, opened 2026-08-25
Allow reshape-equivalent inputs during custom-op execution

ollycassidy13:
# Allow reshape-equivalent inputs during custom-op execution

## Why

Fused reshapes can leave execution-context tensors with a different logical shape from the hardware operator's stream shape, despite having the same number of elements. Strict shape checks caused otherwise valid simulations to fail.

## How

Temporarily reshape compatible inputs for HLS OuterShuffle, RTL InnerShuffle and RTL elementwise execution, then restore the original execution context.

### DIFF
--- a/src/finn/custom_op/fpgadataflow/hls/outer_shuffle_hls.py
+++ b/src/finn/custom_op/fpgadataflow/hls/outer_shuffle_hls.py
@@ -121,7 +121,30 @@ class OuterShuffle_hls(OuterShuffle, HLSBackend):
         ]
 
     def execute_node(self, context, graph):
-        HLSBackend.execute_node(self, context, graph)
+        node = self.onnx_node
+        input_name = node.input[0]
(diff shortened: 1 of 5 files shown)


auphelia:
Reshape-equivalent execution views (Shuffle sites)
Looks reasonable to me. My understanding is that this was a loud fail rather than a silent correctness issue. Is that correct?
One suggestion: instead of a generic reshape, I’d prefer if we use the tracked in_shape / transpose_in_shape attributes since they capture the original execution view explicitly.

Reshape-equivalent execution views (Elementwise RTL)
I don’t really understand this one yet. Unlike the shuffle case, I don’t see where the shape provenance comes from for a generic elementwise op. Replacing the shape assert with a reshape feels like it could hide real issues. What was the actual mismatch you were seeing, and why is a reshape the right fix rather than checking against an expected shape?

PR #1673, opened 2026-08-26
[Builder] Check verify_steps prerequisites are actually run

### LINKED ISSUE #1428: Check the prerequisite steps for the verification flow before executing
## Prerequisites
None

### New behavior
The verification flow should only be started if all necessary transformations ran before. Therefore, checks could be added to ensure that when running FINN in verification mode, all necessary transformations (e.g. create stitched ip) have been run before.

### Motivation
To avoid errors when editing the build flow in the build configuration while having verification mode enabled, make sure that still all steps required to run the verification flow are executed before.

### Parts of FINN affected
Verification Flow

mumallaeng:
Fix #1428.

## Problem

Each `VerificationStepType` entry (`verify_steps`) is only ever triggered from inside one specific step - `stitched_ip_rtlsim` from `step_create_stitched_ip`, `node_by_node_rtlsim` from `step_hw_ipgen`, and so on. If the configured `steps`/`start_step`/`stop_step` excludes that step (e.g. `stop_step` set before `phase_build_hardware` while `verify_steps` still asks for `node_by_node_rtlsim`), the verification is silently never run. The build reports success with no indication that the requested check never happened - exactly the failure mode #1428 describes.

## Fix

Added a `verify_step_prereq` check to `run_all_config_checks` in `build_dataflow_checks.py`. It resolves the actual list of steps that will run via `resolve_build_steps` (the same function `build_dataflow.py` uses internally) and flags any `verify_steps` entry whose owning phase or fine-grained step isn't in that resolved list.

`resolve_build_steps` is imported lazily via `importlib.import_module` rather than a top-level import, since `build_dataflow.py` imports `build_dataflow_checks.py` - a top-level import back would be circular. `.ruff.toml` enables `PLC` (including `PLC0415`, import-outside-top-level), so a plain function-local `from ... import ...` would fail lint; `importlib.import_module` isn't an `Import`/`ImportFrom` AST node, so it isn't flagged.

**Update**: per review feedback, this check is now a WARNING rather than an ERROR (a custom step/flow handling its own verification shouldn't get blocked), and the `(phase, step)` prerequisite mapping now lives next to `VerificationStepType` itself, as `verify_step_prereqs` in `build_dataflow_config.py`, instead of inline in `run_all_config_checks`.

## Testing

Ran the full `test_build_dataflow_checks.py` suite locally (`setup-local.sh`-style venv, no Docker/Vivado) - all 15 tests pass, including 3 new ones: a `verify_steps` entry whose phase is excluded by `stop_step` (errors), one whose phase is included (no error), and a custom `steps` list naming the fine-grained step directly instead of its phase (also no error, since the check accepts either). `isort`/`black`/`ruff` (pinned versions from `.pre-commit-config.yaml`) pass on both changed files.

### DIFF
--- a/src/finn/builder/build_dataflow_checks.py
+++ b/src/finn/builder/build_dataflow_checks.py
@@ -3,6 +3,7 @@
 
 """Configuration check system for FINN builds - catches incompatibilities early."""
 
+import importlib
 import json
 import os
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


auphelia:
HI @mumallaeng, 
Thanks for your patience and for picking this up!

A few suggestions:

1. Change ERROR → WARNING: Users with custom steps that handle verification themselves would get false positives and have to mute all assertions. A warning alerts without blocking.
2. Co-locate prerequisite with enum: Instead of a separate verify_step_prereqs dict, embed the minimum required step in VerificationStepType itself (https://github.com/Xilinx/finn/blob/dev/src/finn/builder/build_dataflow_config.py#L81):
```
class VerificationStepType(Enum):
      FOLDED_HLS_CPPSIM = ("folded_hls_cppsim", "step_minimize_bit_width")
      STITCHED_IP_RTLSIM = ("stitched_ip_rtlsim", "step_create_stitched_ip")
```
3. The check would then verify that all default steps up to and including that step are in the resolved list. This keeps the mapping maintainable, adding a new verification type requires defining its prerequisite.
4. Future: runtime tracking: For custom flows, the most reliable check would be tracking at the end of the build whether verify_step() was actually called for each requested verification. But that could be a follow-up.

mumallaeng:
@auphelia Thanks for the feedback!

1. Downgraded verify_step_prereq to WARNING, so a custom step/flow that handles verification itself won't get a false-positive build-blocking error.
2. Moved the (phase, step) prerequisite for each VerificationStepType out of run_all_config_checks and next to the enum itself, as verify_step_prereqs in build_dataflow_config.py, so a new verification type's prerequisite is defined right where the type is.

One deviation from your snippet: I kept the phase name alongside the fine-grained step rather than dropping to just the step. For a default build (steps=None), resolve_build_steps only ever resolves to the phase-level functions - the fine-grained step names never appear in that list - so a step-only check would false-positive on every default build using verify_steps. Happy to adjust if you had something else in mind.

auphelia (approved):
LGTM! Thanks @mumallaeng for addressing this silent failure mode. Catching misconfigured verify_steps early will save users from builds that look successful but skip the verification they intended.

One minor suggestion for future consideration: could add a test/assertion ensuring all `VerificationStepType` members have entries in `verify_step_prereqs` to catch omissions when new verification types are added.

PR #1675, opened 2026-08-26
Support asymmetric Quant activation conversion

ollycassidy13:
# Support asymmetric Quant activation conversion

## Why

QONNX-to-FINN activation conversion rejected non-zero zero points and could produce incorrect intermediate datatypes for signed, unsigned and narrow quantizers.

## How

Account for scalar and per-channel zero points when deriving thresholds and biases, preserve the unsigned intermediate level range, handle round-to-even boundaries and select a datatype that represents every output level.

### DIFF
--- a/src/finn/transformation/qonnx/qonnx_activation_handlers.py
+++ b/src/finn/transformation/qonnx/qonnx_activation_handlers.py
@@ -179,7 +179,19 @@ class QuantActBaseHandler(ABC):
             mt_inst.set_nodeattr("out_dtype", out_dtype)
         else:
             # Set datatype
-            mt_inst.set_nodeattr("out_dtype", out_dtype)
+            # Signed Quant identity conversion inserts its negative offset as a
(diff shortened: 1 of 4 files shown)


auphelia:
My main concern here is not the arithmetic itself, but why we should lift the existing zero_point == 0 contract at all. FINN already has ExtractQuantScaleZeroPt to rewrite asymmetric Quant nodes into a form that the existing symmetric MultiThreshold conversion can handle. Have you tried that standard extract-and-streamline path on your model? If so, where does it fail or leave operations that cannot be absorbed? Unless there is a concrete reason why that path cannot work, I would prefer to keep the current loud rejection and address the issue through the existing graph rewrite or in the export. Teaching several threshold-conversion paths about zero-points introduces additional silent-offset risk and could leave the existing guards inconsistent. If the in-threshold approach is genuinely required, we would need a clear justification and comprehensive tests covering the affected quantization variants.

PR #1678, opened 2026-08-27
FIFO Generation Characterization Improvements

ollycassidy13:
# FIFO Generation Characterization Improvements

## Why

FIFO generation and characterization could hang, derive incorrect depths for multi-input graphs or generate FIFOs that could not be simulated reliably. Large MLO designs also made characterization impractically slow, while Vivado could intermittently fail to discover generated FIFO modules.

## What

- Match producer and consumer characteristics using the connected tensor ports.
- Supply auxiliary weight streams when characterizing dynamic and external MVAUs.
- Skip closed `FINNLoop` nodes and apply safe boundary FIFO depths.
- Bound and accelerate characterization, with compressed sidecars and linear trace storage.
- Keep oversized FIFOs in RTL and split them into simulatable depth-limited chunks.
- Deduplicate shared RTL FIFO sources and retry transient Vivado module discovery.
- Use RTL FIFO copies during node-by-node simulation when synthesis uses Vivado FIFO implementations.

### DIFF
--- a/finn_xsi/finn_xsi/sim_engine.py
+++ b/finn_xsi/finn_xsi/sim_engine.py
@@ -250,11 +250,13 @@ class SimEngine:
             def __init__(self, sim, stream):
                 self.vld = sim.get_bus_port(stream, "tvalid")
                 self.rdy = sim.get_bus_port(stream, "tready")
-                self.trace = ""
+                self.trace = bytearray()
 
             def __call__(self, sim):
-                self.trace += (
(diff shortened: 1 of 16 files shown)


preusser-amd:
I'd strongly encourage queuing this behind #1658.

auphelia:
Thanks for the PR @ollycassidy13 , can you please have a look if those changes are still required considering the merge of https://github.com/Xilinx/finn/pull/1658 ?

PR #1679, opened 2026-08-28
FIFO Sizing Counter Overflow Fix

mrontio:
FIFO Sizing uses a counter `MaxCount` which must be sized in the HDL. 
This was previously done based on the depth of the inserted FIFO: https://github.com/Xilinx/finn/blob/a2c05ab445c140404145aeb4ef8b4b1fcb1e2627/src/finn/custom_op/fpgadataflow/rtl/streamingfifo_rtl.py#L91

When building siglip, an unusually large residual path has overflowed this counter and caused weeks of delay to find the stall.
I set the COUNT_WIDTH to default 32 bits which should make overflow impractical.


auphelia:
Thanks for tracking this one down, @mrontio! A sizing FIFO that only deadlocks at runtime is exactly the kind of bug that costs weeks to trace. Thomas and I applied a few more changes on top of your PR, but only small things.

PR #1684, opened 2026-09-02
Adding credit-based flow control to read issuing.

preusser-amd:
The added credit-based flow-control ensures that there is sufficient buffer space to consume all replies for all memory read requests in flight. It avoids using the path through the memory controller for backpressuring the read request generation.


d-kor:
Credits getting released on the AXI stream, what happens if the LWB is backpressuring, there are pipeline cycles in between? 
Maybe increment credit on the actual consumer handshake instead of r_hsk? Credit on the actual buffer you are protecting.

d-kor (approved):
All good. Credit system upstream from LWB, this is fine, just extra space for beats.

preusser-amd:
Obsolete by #1702.

PR #1688, opened 2026-09-05
Fix estimate-only build crash in OuterShuffle when Vivado is not configured

### LINKED ISSUE #1677: Estimate-only build crashes: outer_shuffle.py's get_exp_cycles() does `re.search()` on unset XILINX_VIVADO
## Summary

An estimate-only `build_dataflow` run (no Vivado installed/configured — using the local,
non-Docker `setup-local.sh --skip-xsi` path) crashes with `TypeError: expected string or
bytes-like object` when the model produces an `OuterShuffle`/`InnerShuffle` hardware layer.
The docs and general guidance say estimate reports don't require Vivado, but this code path
assumes `XILINX_VIVADO` is always set.

## Environment

- Branch: `dev`, commit `aac6c65b5baadd3a54f3dd9b819ecb745a01f4a7`
- Install method: `./setup-local.sh --skip-xsi` (no Docker, no Vivado/Vitis HLS)
- `FINN_XILINX_PATH` / `FINN_XILINX_VERSION` / `XILINX_VIVADO`: unset
- OS: Ubuntu 22.04 (WSL2), Python 3.10.12

## Steps to reproduce

1. Build any model that produces a `Shuffle` hardware layer (e.g. one whose input needs an
   internal NCHW/NHWC-style data-layout adaptation during `convert_to_hw`).
2. Run `build_dataflow_cfg` with `generate_outputs=[DataflowOutputType.ESTIMATE_REPORTS]` and
   `steps=estimate_only_dataflow_steps`, with no Vivado configured.
3. The build fails inside `step_generate_estimate_reports`.

## Traceback
Traceback (most recent call last):
File "finn/builder/build_dataflow.py", line 171, in _run_build_steps
model = transform_step(model, cfg)
File "finn/builder/build_dataflow_phases.py", line 237, in phase_optimize_hardware
model = _execute_step(step_generate_estimate_reports, model, cfg)
File "finn/builder/build_dataflow_phases.py", line 80, in _execute_step
model = step_fn(model, cfg)
File "finn/builder/build_dataflow_steps.py", line 762, in step_generate_estimate_reports
estimate_layer_cycles = model.analysis(exp_cycles_per_layer)
File "qonnx/core/modelwrapper.py", line 164, in analysis
return analysis_fxn(self)
File "finn/analysis/fpgadataflow/exp_cycles_per_layer.py", line 47, in exp_cycles_per_layer
cycle_dict[node.name] = int(inst.get_exp_cycles())
File "finn/custom_op/fpgadataflow/outer_shuffle.py", line 190, in get_exp_cycles
match = re.search(r"\b(20\d{2}).(1|2)\b", vivado_path)
File "/usr/lib/python3.10/re.py", line 200, in search
return _compile(pattern, flags).search(string)
TypeError: expected string or bytes-like object

## Root cause

`custom_op/fpgadataflow/outer_shuffle.py` (~line 189):

```python
vivado_path = os.environ.get("XILINX_VIVADO")
match = re.search(r"\b(20\d{2})\.(1|2)\b", vivado_path)
year, minor = int(match.group(1)), int(match.group(2))
os.environ.get("XILINX_VIVADO") returns None when unset, which re.search can't accept
as its string argument. The purpose of this check is just to pick between two analytical
pipeline-latency models (pre/post Vivado-2024.2 URAM read-latency behavior) — it doesn't
actually invoke Vivado — so there's no reason it should require Vivado to be installed at all
for an estimate-only build.

A sibling helper, finn/util/basic.py's get_vivado_version(), handles exactly this case
correctly:


path = os.environ.get("XILINX_VIVADO", "")
match = re.search(r"\b(20\d{2})\.(1|2)\b", path)
return (int(match.group(1)), int(match.group(2))) if match else None
custom_op/fpgadataflow/hlsbackend.py (~line 256, compile_singlenode_code) has the same
unguarded pattern, though that path is only reached during cppsim compilation rather than
plain estimate-report generation.

##Suggested fix
Have outer_shuffle.py::get_exp_cycles() (and hlsbackend.py::compile_singlenode_code) use
finn.util.basic.get_vivado_version() instead of inlining the same regex without a default,
and fall back to a sensible default pipeline model (e.g. assume II=1, the pre-2024.2
behavior) when no Vivado version is available, rather than crashing.

##Workaround
Setting XILINX_VIVADO to any string matching \b20\d{2}\.(1|2)\b (e.g.
export XILINX_VIVADO=/opt/Xilinx/Vivado/2022.2, without an actual Vivado install at that
path) avoids the crash for estimate-only builds.

### PR DESCRIPTION
**Pull request by chiliec**
## What

Closes #1677.

An estimate-only `build_dataflow` run (no Vivado configured) crashes with
`TypeError: expected string or bytes-like object, got 'NoneType'` when the model
produces an `OuterShuffle` layer.

## Fix

`OuterShuffle.get_exp_cycles()` did:

```python
vivado_path = os.environ.get("XILINX_VIVADO")
match = re.search(r"\b(20\d{2})\.(1|2)\b", vivado_path)
year, minor = int(match.group(1)), int(match.group(2))
```

`os.environ.get("XILINX_VIVADO")` is `None` when Vivado is not configured, which
`re.search` can't accept. This version check only selects between two analytical
pipeline-latency models (pre/post Vivado-2024.2 URAM read-latency) — it doesn't
invoke Vivado — so it shouldn't require Vivado for an estimate-only build.

The change reuses the existing `finn.util.basic.get_vivado_version()` helper
(which already handles the unset case by returning `None`) and falls back to the
pre-2024.2 pipeline model when no Vivado version is available.

## Tests

Added `test_outer_shuffle_exp_cycles_without_vivado` in
`tests/fpgadataflow/test_fpgadataflow_shuffle.py`: it builds a model that
decomposes into an `OuterShuffle_hls` node, unsets `XILINX_VIVADO`, and asserts
`get_exp_cycles()` returns a positive int instead of crashing.

## Validation

Ran locally in a CPU venv (no Vivado installed):

```
$ pytest tests/fpgadataflow/test_fpgadataflow_shuffle.py::test_outer_shuffle_exp_cycles_without_vivado
1 passed

# RED->GREEN: reverting the source fix (keeping the test) makes it fail
1 failed   (TypeError: expected string or bytes-like object, got 'NoneType')
```

Also confirmed `get_exp_cycles()` returns the same cycle count with
`XILINX_VIVADO` unset, `.../2022.2`, and `.../2024.2`, and that the other
non-Vivado shuffle tests in that file still pass. `pre-commit` hooks
(isort/black/ruff) are clean on the changed files.

Note: the sibling unguarded pattern in `hlsbackend.py::compile_singlenode_code`
(cppsim path) is left untouched here — this PR is scoped to the reported
estimate-only crash. Happy to address that in a follow-up if you'd like.


auphelia:
Hi @chiliec, thank you for the PR! Sorry for the delay, I am currently travelling for work. Thanks for addressing the issue, I think overall this is the right approach, but I believe the default behaviour should be to assume 2024.2 or higher, because that is what we recommend now when using FINN. I think it would also be good to give the user a warning that Shuffle estimations are version-dependant and a Vivado version was assumed.

chiliec:
Thanks @auphelia, and no rush — hope the travel is going well. Both points done in 586afca6e:

- **Default is now 2024.2+**: when `XILINX_VIVADO` is unset the estimate uses the 2024.2+ pipeline-II model (BRAM → II=1, URAM-depth buffers → II=3) rather than falling back to the pre-2024.2 II=1 path.
- **Warning**: a `warnings.warn` fires in that case — `"<node>: XILINX_VIVADO is not set; OuterShuffle cycle estimates are Vivado-version dependent, assuming Vivado 2024.2 or newer."`

The test (`test_outer_shuffle_exp_cycles_without_vivado`) now asserts the warning is emitted and that the no-Vivado estimate is identical to the estimate with a 2024.2 install configured. Configured-Vivado paths are unchanged (no warning, same II selection as before). black/isort/flake8 clean on the diff.

auphelia (approved):
Thanks a lot for the PR @chiliec ! Looks good to me.

PR #1690, opened 2026-09-10
SigLIP Transformer Example

ollycassidy13:
# Add a SigLIP ImageNet example for VCK190

## Why

This adds a reproducible FINN flow for the image tower of `google/siglip2-base-patch16-224`. The example validates the W6A7 QV-LSQ model handoff, rolls the repeated transformer blocks into a `FINNLoop`, and provides the folding and specialization needed to generate a routed VCK190 design.

## What changed

- Add a phase-based CLI covering SigLIP preparation, optimization, hardware conversion, stitched-IP verification, and routed OOC synthesis.
- Add model/report validation, deterministic reference-vector generation, and end-to-end Python and RTL tests.
- Roll the repeated transformer layers into a `FINNLoop` and extend the conversion/RTL support needed by the image tower, including asymmetric activations, residual stream duplication, reshapes, and external MVAU weights.
- Harden FIFO sizing, weight-memory banking and XSI simulation so the large finite-FIFO design remains live and fits on VCK190.
- Add the reviewed W6A7 folding and specialization profiles and document the measured ImageNet accuracy, timing, and resource results.

## SigLIP Preparation

Quantized SigLIP models were exported from a flow following [Quantization Variation](https://arxiv.org/pdf/2307.00331).


auphelia:
Thanks for opening these PRs and for all the work that went into bringing TinyDeiT and SigLIP support to FINN, @ollycassidy13.

I've converted these PRs to draft for now. While they're valuable as a record of your work and contributions that helped drive transformer support in FINN, they're not something we'd want other users to use as examples or a starting point today. The TinyDeiT PR currently contains merge conflicts, and both PRs include a substantial amount of exploratory code and implementation attempts that were useful during development but would likely be confusing for users trying to understand the recommended workflow.

Most of the functionality developed through this work has since been merged into FINN through a series of smaller, focused PRs that introduced the required infrastructure and features in a cleaner and more maintainable way (thanks for those PRs @ollycassidy13!). As a result, these PRs are no longer representative of how transformer support should be added to FINN today.

I still think it's useful to keep them around for now as they document your contributions and the development journey that ultimately enabled transformer support in FINN. We also plan to add a clean end-to-end transformer example in the near future, which should serve as the recommended reference for users interested in experimenting with transformers in FINN.

Issue #1694, opened 2026-09-17
Pool (MaxPool): output datatype does not follow the input, `get_output_datatype` asserts `odt == idt`


fpjentzsch:
## Summary

`Pool.get_output_datatype` reads both `InputDataType` and `OutputDataType` from node attributes and, for `Function == "MaxPool"`, asserts that they are equal. `Pool.infer_node_datatype` re-annotates the output tensor from the `OutputDataType` attribute only ("data type stays the same") and does not update either attribute from the actual input tensor annotation.

Both attributes are set once by `InferPool` (`odt = idt` for MaxPool). If the producer's datatype changes afterwards - the normal case when the pool follows a layer whose output is narrowed by `MinimizeAccumulatorWidth`, e.g. a MVAU without activation - the node keeps the stale, wider types:

* the assertion cannot fire (both attributes are stale and equal), but the HLS code is generated for the stale `InputDataType` while the AXI stream carries the producer's narrower type, and the consumer sees a stale output annotation;
* if `infer_node_datatype` is changed to track the input (as it should, and as we do in FINN+), the assertion fires instead of the output simply following the input.

## Suggested fix

For `MaxPool`, `infer_node_datatype` should set both `InputDataType` and `OutputDataType` from the input tensor annotation (output = input by definition) instead of asserting equality. For the accumulating variants (`QuantAvgPool`, and in FINN+ `AccPool`/`AvgPool` converted from `ReduceSum`/`ReduceMean`) the accumulator/output widths should be re-derived from the current input type in a `minimize_accumulator_width` (see the separate issue: https://github.com/Xilinx/finn/issues/1695); for `ReduceMean`/`AvgPool` in particular the `AccumBits` attribute is fixed at conversion time from the (placeholder) input type and never updated.

## How we addressed it in FINN+ for now

`Pool.infer_node_datatype` updates `InputDataType` from the input tensor and, for `MaxPool`, passes it through to `OutputDataType`; `Pool.minimize_accumulator_width` re-derives `AccumBits`/`OutputDataType` for `AccPool`/`AvgPool`.

auphelia:
Thanks @fpjentzsch and @mumallaeng! Closing this issue since the PR is merged.

PR #1697, opened 2026-09-19
[Pool] Track producer's datatype in MaxPool's infer_node_datatype

mumallaeng:
Fixes #1694.

## Problem

`Pool.infer_node_datatype` only re-wrote the output tensor annotation from the stored `OutputDataType` attribute. `InputDataType` was never synced against the actual input tensor annotation. If the producer's output datatype narrowed after conversion (e.g. via `MinimizeAccumulatorWidth` on an upstream node), `Pool`'s own `InputDataType`/`OutputDataType` stayed stale and equal to each other, so `get_output_datatype`'s `assert odt == idt` never caught it - but the HLS backend still generated its interface width from the stale `InputDataType` while the AXI stream carried the producer's narrower type.

## Fix

`infer_node_datatype` now reads the live input tensor annotation and writes it into `InputDataType` (same pattern `MatrixVectorActivation` already uses), warning if it changed. For `MaxPool`, `OutputDataType` is also set to match, since output = input by definition.

`QuantAvgPool` is deliberately left untouched - see the commit message for why syncing its `InputDataType` too would make the Python (`execute_node`) and already-generated HLS results diverge.

## Testing

Added `test_pool_maxpool_infer_datatype_follows_input` (`tests/fpgadataflow/test_convert_to_hw_pool_batch.py`), pure Python/qonnx, no Vivado needed. Confirmed it fails on the pre-fix code (`assert UINT8 == UINT4`) and passes after the fix. Also ran the full non-Vivado-marked test selection in that file and confirmed both `pool.py` and `convert_to_hw_layers.py` still import cleanly. `isort`/`black`/`ruff` (pinned versions from `.pre-commit-config.yaml`) and `python -m py_compile` pass on both changed files.


auphelia (approved):
Thanks @mumallaeng ! Looks good to me.

PR #1698, opened 2026-09-19
[SWG] Fix depthwise deadlock at feature-map end when input is slower than window rate

fpjentzsch:
## Problem

The RTL sliding-window generator (`swg_template_default.sv` and `swg_template_default_dynamic.sv`) deadlocks on the **last element of a feature map** in **depthwise** mode whenever its input stream arrives no faster than the SWG's own window rate.

`First_elem_next_window` has the same width as the other element counters (`clog2(LAST_READ_ELEM+1)+1` bits). At the first fetch of a feature map's last window it is advanced by `TAIL_INCR_LAST = buffer_min_size - 1`. In depthwise mode the tail increments of the preceding windows only accumulate to `channel_factor - 1`, so the register legitimately ends up at `channel_factor - 1 + buffer_min_size - 1`, which exceeds `LAST_READ_ELEM`. `read_cmd` compares it via `$signed()`, where the wrapped value turns negative (e.g. 8702 → −7682 in a 14-bit register), so the condition `Newest_buffered_elem - (BUF_ELEM_TOTAL-1) < First_elem_next_window` is false and the SWG never asserts `TREADY` again.

The condition is only reached if the last window *starts* before the last input element has been read — i.e. when the producing layer is slower than or exactly as fast as the SWG (one window per `ELEM_PER_WINDOW` cycles). If the producer is faster, all elements are already buffered at that point and nothing happens, which is presumably why it went unnoticed. Non-depthwise SWGs are unaffected: their accumulated tails land exactly on `LAST_WRITE_ELEM`.

## How it surfaced

ResNet-18 (CIFAR-100) build with a global average pool implemented as depthwise SWG (4×4 kernel over 4×4, 512 channels, SIMD 1, i.e. one 16-element window per 16 cycles) followed by `Pool`: with the preceding MVAU delivering one element per 16 cycles the FIFO-sizing RTL simulation never completed a single image; at one per 12 cycles it worked. Transaction counters pinned the stall to exactly the 8192nd element of the first image at the SWG's input.

## Fix

Widen `First_elem_next_window` by one bit in both templates. No functional change for values within the old range.

## Verification

xsim testbench replaying the input pattern against the generated module (rendered from the `dev` template with the parameters above); `in`/`out` are accepted/emitted elements for 3 feature maps:

| cycles/element | `dev` | fixed |
|---|---|---|
| 12 | 24576 / 24576 | 24576 / 24576 |
| 16 | **8191 / 8191 (stall)** | 24576 / 24576 |
| 32 | **8191 / 8191 (stall)** | 24576 / 24576 |

Other depthwise variants (3×3 stride 1, 2×2 stride 2, SIMD 4) show the same threshold at `ELEM_PER_WINDOW` cycles per element on `dev` and pass with the fix; 3×3 non-depthwise SWGs pass at all rates in both versions. Output order was checked in the testbench.

Same fix in finn-plus: https://github.com/eki-project/finn-plus/pull/264


preusser-amd:
Thanks so much, @fpjentzsch!
I've found another issue with pointer initialization and reviewed and annotated all range allocations exploiting the implied width reduction potential. Please, review.

fpjentzsch:
> Thanks so much, @fpjentzsch! I've found another issue with pointer initialization and reviewed and annotated all range allocations exploiting the implied width reduction potential. Please, review.

Thanks @preusser-amd, looks good and all tests pass on our end.

auphelia (approved):
Thanks @fpjentzsch and @preusser! I have added two small test cases that check for the bug that is resolved with this PR. When the GHAs are passing, this will be merged.

